PR #8, opened 2010-09-14
Add support for HTML comments

felixge:
Previously I NERDCommenter tried to use C-style comments (/\* */) for my HTML files, this commit fixes this.


othree:
One possible reason is writing in template file, ex: Jinja2
<https://github.com/Glench/Vim-Jinja2-Syntax/pull/13/files>

PR #119, opened 2013-06-07
Added Idris delimiters

raichoo:
Added delimiters for the dependently typed function programming language Idris
(http://www.idris-lang.org)


alerque:

**About this PR:** I was going to merge this, but there was another PR with the same addition that was both earlier and alphabetically sorted correctly. It did however use different delimiters by default so if there is something to fix there let me know. Please consider installing the plugin from `alerque/nerdcommenter` for this and (so far) over two dozen other updates.

alerque:
Change of plans. Thanks to scrooloose who has added me as a maintainer to this repo, I'll be going through and taking care of business here instead of in a fork.  Per comments above, I've chosen to merge PR #139 in leu of this one. If there is a problem with that or a case to make that the default comment style should be different, please comment on one of these PRs or open a new one that fixes it and make the case for consideration.

Even if we couldn't use it in exactly this form thanks for your contribution @raichoo.

PR #123, opened 2013-08-16
Update Python comment style

mozillazg:
The [PEP8](http://www.python.org/dev/peps/pep-0008/#inline-comments)\(Style Guide for Python Code\) said:

> An inline comment is a comment on the same line as a statement. 
> Inline comments should be separated by at least two spaces from the statement.
>  **They should start with a # and a single space**.

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -303,7 +303,7 @@ let s:delimiterMap = {
     \ 'psf': { 'left': '#' },
     \ 'ptcap': { 'left': '#' },
     \ 'puppet': { 'left': '#' },
-    \ 'python': { 'left': '#' },
+    \ 'python': { 'left': '# ' },
     \ 'radiance': { 'left': '#' },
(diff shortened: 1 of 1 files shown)


alerque:

**About this PR:** I would have merged this PR, but as it stands there are several reasons to think this isn't the right solution. There is already an option for adding spaces after delimiters and it seems better to use that then fight over the best default (personally I think most languages look better with).

alerque:

Per the comments above, I have some reservations about this. I understand why matching the languages own style guide would be a good thing, but the space issue has a config option already so I don't think adding it as part of the delimiter is such a great idea especially when _technically_ it works without as well.

Perhaps if there was a way to trigger the space handling setting by default on a per language basis? How would that sit with folks?

alerque:
As of #183 having been merged (and re-fixed in #239 after I broke it) this is the default behavior so I'm going to close this PR as a duplicate. Please also see #202 for discussion about whether this is actually the right solution or not.

PR #24, opened 2011-02-05
Default alignment option + ability to align at index 0.

ervandew:
I add a couple of features that I wanted for commenting that others may find useful:

1. an option to set the default alignment value used by commands which don't have a specific intended alignment.  Using this I can specify how I want NERDCommenterToggle to align my comments.
2. a new alignment value, 'start', which indicates that the comment characters should be inserted at index 0 (column 1) of the line.  As a personal convention, when I'm temporarily commenting code out, I prefer my comment delimiters to be at the very start of the line so that I can quickly pick out what comments are temporary vs permanent.

In addition to updating the help file to document these new features, please also note one gotcha where using the 'start' alignment with g:NERDRemoveExtraSpaces can result in the undesired removal of spaces.  I thought about adding some code to detect and warn the user in this situation, but decided against it at the time.  Its something worth considering when decided whether these changes are worth merging in.

PR #137, opened 2013-12-04
Add 'c0 to comment in first column

lxsli:
Thanks for looking.


alerque:
Thanks for taking the time to contribute @lxsli. I've recently been added as a maintainer on this project and am trying to work through the backlog of contributions. Unfortunately this doesn't seem to be in shape to be merged at this time. There are some merge conflicts and it's not clear to me that this will work as intended without some tweaking.

Can I ask that you do two things to push this along?

 1. Rebase your fork based on the current master branch and resolve and conflicts.

 2. Review the changes that I pulled in with PR #24 and see how they affect this? It seems like they both accomplish roughly the same thing different ways. Given the large amount of support for that PR I went ahead and merged it, but if there is something significant to be gained from your approach I'd me glad to consider it in addition to the other setting.

Thanks again, and happy commenting.

lxsli:
I'm quite happy with the merged patch, thanks

alerque:
Thanks for reviewing!

PR #140, opened 2014-01-13
add Racket setting

cmonkey:


alerque:

**About this PR:** I have merged a previous PR that added support for Racket. Yours is a little different, and I'd be willing to consider it but I have a hard time believing a language uses the same delimiters for multi-line comments as it does for single line ones. It looks like this may not be correct. If it is please let me know specifically and I'll merge yours as an update to the previous. In the mean time please consider installing the plugin from `alerque/nerdcommenter` for this and (so far) over two dozen other updates.

alerque:

As mentioned above, this language has already been added. If there is some need to the alternate block syntax as well, please comment here to let me know what that is or open a PR that is based on the current master that fixes the existing entry.

Even if we didn't use it as is, thanks for your contribution @cmonkey.

PR #141, opened 2014-02-02
First step to add Only Sexy comments feature.

rendon:
Hi! I'm proposing the following feature. What do you think about it?

I find this plugin very useful but actually the only feature I use
most of the time is the so called "Sexy Comments", however this feature doesn't
behaves exactly as I want. For example:

Default behavior for toggle comments:
```
    //while (true) {
        //Socket client = sock.accept();
        //PrintWriter pout = new PrintWriter(client.getOutputStream(), true);
        //pout.println(new java.util.Date().toString());
        //client.close();
    //}
```

Default behavior for sexy comments:
```
    /*
     *while (true) {
     *    Socket client = sock.accept();
     *    PrintWriter pout = new PrintWriter(client.getOutputStream(), true);
     *    pout.println(new java.util.Date().toString());
     *    client.close();
     *}
     */
```
Desired behavior in both cases:

```
    //while (true) {
    //    Socket client = sock.accept();
    //    PrintWriter pout = new PrintWriter(client.getOutputStream(), true);
    //    pout.println(new java.util.Date().toString());
    //    client.close();
    //}
```
The feature I'm proposing is to create a mode that, when ON, the Toggle function calls
CommentLinesSexy() by default and this function behaves as shown above.

There is a lot of work to do for this feature to work with all the languages
supported, this is only the first step. The tested languages are:

- Python
- C/C++
- Java
- Ruby
- Bash
- VimML

Personally I have disabled all the mapping of this plugin and only have
<leader>c binded to SexyCommentsOnly. Don't worry this isn't part of my
SexyCommentsOnly feature.


alerque:
Thanks for your contribution @rendon. I've been added as a maintainer to this repo and will be going through all the old pull requests. At first bluff this one seems to accomplish much the same thing as #24 that I merged just a minute ago.

Can you give that change a review and let me know where this is heading that's different than that? And where you are at with this development wise?

rendon:
Thanks @alerque for responding, I will try #24 from your repo and I'll let your know if that covers my pull request.

rendon:
Indeed #24  does the same as this PR. You can close this one :).

alerque:
Thanks for looking into this @rendon. I'm closing this now, but let me know if anything turns up that isn't functioning as well as it could and we'll review the implementation differences.

Issue #118, opened 2013-06-06
Support CoffeeScript block comments


adambiggs:
More info at the bottom of [this section](http://coffeescript.org/#strings).

Pretty useful in CoffeeScript since they're included in the output JS, whereas single line comments aren't.

PR #142, opened 2014-02-10
wrong coffee script coment

aemonge:
the "#" comments outputs as comments, not actual comment


alerque:
As noted in comments on this commit, this seems to be the wrong solution. If anybody that knows and can test CoffeeScript wants to submit a PR with the triple `#` block syntax as described here, I'd be glad to consider it for merging upstream. In the mean time I'm going to close this PR.

alerque:
See also #118, and note I've added the triple hash delimiters so you can get block comments using the minimal comment function now. Anything else?

Issue #131, opened 2013-09-27
Support for ocaml.


jordwalke:
Ocaml comments seem to be commented as Java style comments.

alerque:
Per my [comment on PR #151](https://github.com/scrooloose/nerdcommenter/pull/151#issuecomment-221203628), I am not able to reproduce the problem in the original post here. I get a different behavior that changes the nested comment delimiter. @nilsbecker can you confirm that the steps in your comment are still actually a problem for you? Because they work for me and this might have gotten fixed already. If not I need to figure out how #151 fits in.

nilsbecker:
i just checked again the steps in my comment from 2 jun 2014. the exact result is not the same but this is still not working. i am using nerdcommenter as of a few days ago. 

specifically, make a file `whatever.ml` and put the line 
``` let open Reaction in```
in there. then comment one word as per the first step in that post. now comment the line. what is see is that the commented word (the nested comment) is now surrounded by `[> <]` which is wrong. nested comments should be unchanged; ideally uncommenting should dissolve the nested comments from within -- as much as that is possible. it seems not easy to solve with a line-based commenting logic

alerque:
@nilsbecker Please update to nerdcommenter as of a few minutes ago. I've merged over 3 dozen pull requests in the last 24 hours, so it's important that we're having the discussion around the current behavior, not the old. Thanks...

nilsbecker:
ok, i did and it's the same. you should find it as well? the first two steps in the old comment for me give
``` (* let open [> Reaction<] in *) ```
which is not desired.

nilsbecker:
i'm not sure this is what you meant? do i need to check out some other branch for testing #151 ? if so this is a bit much right now.

klaasb:
@alerque Basically the problem was (and from looking at the current code, still is), that NERDCommenter only supports nested comments as a workaround for languages that don't support real nested comment syntax. That workaround renames inner comment delimiters as described by @nilsbecker. My approach in #151 was to make that an attribute of the language definition in the delimiterMap and not use that workaround for those languages at all.

@nilsbecker There seems to be a CommentRegion function now, so even removing inner comments should be doable, if this PR would be merged. 

I'm actually not programming in languages supporting real nesting currently and as I wrote, actually switched to tcomment, but I'll have a look later and fix the merge conflicts.

klaasb:
Fixed the merge conflict, rebased on master and shortly tested Nils' test case.
[...]

PR #151, opened 2014-05-21
Support automatic real nesting of comments without placeholders, if supported by the language.

klaasb:
E.g. for haskell, scheme, scala, lisp, ocaml, sml. Fixes issue #131

Implemented by adding 'nested'/'nestedAlt' keys to the language delimiter defintions.

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -195,7 +195,7 @@ let s:delimiterMap = {
     \ 'groovy': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'gsp': { 'left': '<%--', 'right': '--%>', 'leftAlt': '<!--','rightAlt': '-->'},
     \ 'gtkrc': { 'left': '#' },
-    \ 'haskell': { 'left': '{-','right': '-}', 'leftAlt': '--' },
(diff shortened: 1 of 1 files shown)


alerque:
Hey @klaasb and thanks for taking the time to contribute. I've recently been added as a maintainer on this project and am working my way through the backlog of issues and contributions. I'm having some trouble reviewing this one and was wondering if I could get some help. For one thing with all the other merges I've done there are now some merge conflicts in this branch. Can you rebase it to take care of those?

Also I'm having trouble duplicating the issue that this would solve (#131). The exact procedure explained there behaves correctly for me (by changing the nested comment to an alternate syntax) so I was wondering if there is even still a need for this. If it's a better fix that whatever has already been done I'd love to consider it, but can you at least review it and comment on whether and why it is still needed?

klaasb:
Fixed the merge commits, rebased against master and added nesting attributes for Racket, as I saw that scrolling past. However, more languages with nesting might have been added since I wrote this PR.

alerque:
Thanks a bunch @klaasb. This seems to work in my testing so I've gone ahead and merged it. I guess now we wait for people who actually use nested languages to tell us if we broke something. Thanks for the help.

PR #152, opened 2014-05-28
Change Python comment style to match PEP8 standard

jturmel:


jonafato:
As of the merging of #183, this is fixed.

alerque:
This has long since been "fixed". There is some discussion _vis-à-vis_ #202 whether this was the right fix or not, I'm closing this PR as a duplicate. Thanks for taking the time to contribute @jturmel.

PR #153, opened 2014-06-20
add space in c /*  */ comment

idning:
/*nothing to send*/

=================================>

            /* nothing to send */


alerque:
Thanks for taking the time to contribute @idning. As not all people have the same preference though I'm afraid I won't  be able to merge this as is. Fear not though, because this is already setup as a user defined preference. If you prefer spaces inside your comment delimiters, please add this to your rc file:

```
let g:NERDSpaceDelims = 1
```

Issue #145, opened 2014-03-13
NERD_<filetype>_alt_style doesn't work


Wyatts:
It's documented but, from what I can tell of the source, it's not even implemented.

alerque:
This seems to be a duplicate of #33. I'll review the PR as soon as I can get to it. I've only just been added as a maintainer and am working through the issue backlog.

PR #155, opened 2014-07-25
Implement functionality to chose the alternative comment style

saep:
This somewhat fixes Issue #145.

I have no experience in VimL whatsoever, but this seems to be the proper way to respect the NERD_\<filetype\>_alt_style feature. 

If you agree that this is a valid fix, I'd have no problem creating another commit that updates the documentation.

### DIFF
--- a/doc/NERD_commenter.txt
+++ b/doc/NERD_commenter.txt
@@ -714,11 +714,11 @@ again.
 
 If you want the NERD commenter to use the alternative delimiters for a
 specific filetype by default then put a line of this form into your vimrc: >
-    let NERD_<filetype>_alt_style=1
+    let g:NERD_<filetype>_alt_style=1
 <
 Example: java uses // style comments by default, but you want it to default to
 /* */ style comments instead. You would put this line in your vimrc: >
-    let NERD_java_alt_style=1
+    let g:NERD_java_alt_style=1
 <
 
 See |NERDComAltDelim| for switching commenting styles at runtime.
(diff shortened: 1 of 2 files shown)


luisgerhorst:
:+1: Thanks, works fine.

luisgerhorst:
I have a problem with using this version with split windows.

I have set `let g:NERD_erlang_alt_style=1 ` in my `.vimrc`. When I now split my window with an erlang file open using `CTRL-w s` and open a file in another language (e.g. my `.vimrc`) in that buffer, and then switch back to the erlang file using `CTRL-w w` and use `<LEADER>c<space>` to comment out a line it uses the normal comment style again and not the alternate one. Note then I don't have this problem when I simply click on the erlang buffer to switch back to it in MacVim.

saep:
I actually have the same problem by switching between LaTeX and Haskell. Since I almost exclusively use the alternative delimiters and I don't want to spend more time on this, this pull request will probably remain kind of broken.

alerque:
Thanks for taking the time to contribute @seap! I've recently been added as a maintainer on this project and have been going through the backlog of  issues and contributions. I'd really like to see this included as quite a number of people have noted that it doesn't work as documented.

In looking through this it looks like there are a few issues that need to be touched up before it's ready for prime time. Notably the state of regular vs. alternate delimiters should probably be stored in a buffer specific variable (`b:...` instead of the global `g:...`) so that it doesn't get messed up while switching panes, tabs, etc. Are you around at all and are you interested in patching this up?

saep:
Feel free to adjust and use this patch in an way you like. I am not using this plugin anymore and I don't like commiting untested code.

PR #156, opened 2014-08-04
added Coq support

knight-erraunt:


alerque:
Thanks for your contribution @knight-erraunt. I have cherry-picked this commit rather than directly merged it so I could drop the redundant alternate comment syntax. I don't see the point of having identical delimiters in the alternate registers. If I'm missing something here please feel free to let me know.

Issue #33, opened 2011-03-24
"Default delimiter customisation" doesn't work.


zhou13:
According to document, 
add
    let NERD_cpp_alt_style=1
    let NERD_c_alt_style=1
to vimrc should switching commenting style. But it doesn't. And I doesn't see any related code in the script.

alerque:
I've recently been added as a maintainer on this project and am working through the issue backlog. It does look like this feature was never implemented and only the documentation exists. I don't have time to do a bunch of dev work on this—I'm mostly just coordinating all the contributions—but if somebody would like to actually set this up I'd be glad to review a PR that fixes this. Otherwise the thing to do might be to delete the bogus documentation. Anybody?

alerque:
See PR #155 for a (partial?) implementation of this.

alerque:
With many thanks to @saep for actually implementing this, I have merged the PR that adds this feature and made a few tweaks. Please note that I changed the actual preference key format to be more like the other values. For example to always default to the alternates for java:

```vim
let g:NERDAltDelims_java = 1
```

Note this is documented in help and now in the README as well. Let be know if anybody has trouble with the feature.

PR #160, opened 2014-09-19
modify c comment action

kdurant:
/* */ to //

I think ‘//’ is better than ‘/* */’ with single-line comments

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -102,7 +102,7 @@ let s:delimiterMap = {
     \ 'caos': { 'left': '*' },
     \ 'calibre': { 'left': '//' },
     \ 'catalog': { 'left': '--', 'right': '--' },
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for taking the time to contribute @kdurant. This is very much a matter of personal preference and I don't think we can just flip flop this for everybody. This is complicated by the fact that the default 'c' style is used for unknown file types which often use c style comments. However they may not work with line-wise delimiters so this could break a lot of things.

Please follow progress on #33 as there seems to be a call for defaulting to alternate delimiters for some file types. Once that works you should be able to choose the alternates by default. In the mean time you can just set custom delimiters for the file type yourself:

```vim
let g:NERDCustomDelimiters = {
  \ 'c': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
\ }
```

PR #161, opened 2014-09-27
Add ability to create a line of comment characters

ElPiloto:
Add functionality to insert a line from comment characters and to toggle
back to a blank line.

This requires the following modifications:

Simple function to insert the line of comment characters.

Modify UncommentNormalLine() to detect a line of pure comment characters
and remove all of them (i.e. toggle).

Create default mapping (<Leader>c_)

Add option NERDNumCommentCharsHR to specify the width of the line made
from comment characters.

Update help documentation for new functionality, default key mapping,
and new option.

### DIFF
--- a/doc/NERD_commenter.txt
+++ b/doc/NERD_commenter.txt
@@ -28,6 +28,7 @@ CONTENTS                                               *NERDCommenterContents*
             3.2.11 Use alternate delims map...|NERDComAltDelim|
             3.2.12 Comment aligned maps.......|NERDComAlignedComment|
             3.2.13 Uncomment line map.........|NERDComUncommentLine|
+            3.2.14 Comment horiz. line........|NERDComCommentHorizontalRule|
         3.4 Sexy Comments.....................|NERDComSexyComments|
         3.5 The NERDComment function..........|NERDComNERDComment|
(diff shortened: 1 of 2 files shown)


alerque:
Thanks for taking the time to contribute @ElPiloto. I've recently been added as a maintainer  to this project and am  going through the backlog of contributions. This is a neat idea but it's a bit rough around the edges and I don't think I can merge it at this time.

I played around with it a little bit including fixing the merge conflicts then cleaning up the variable names. You can see the result in my [horizontal-rules](https://github.com/alerque/nerdcommenter/tree/horizontal-rule) branch. If you'd like to keep working on this I'd invite you to start from that code, but there are still some things missing:

* Non homogeneous delimiters are messy. This works great for single character delimiters, but when the delimiter is something like `<%#` (ruby) the result is a mess.
* It doesn't properly uncomment lines.
* It isn't taking into account that there might be some characters on the line already and blinding throws 72 new characters into the buffer even if part of the line was drawn.

I'd like to see something more like the sexy comment format that figures out and intermediary character and fills with that rather than just repetition of the opening delimiter. Would it be possible for you to work on these things.?

PR #168, opened 2014-11-12
Add ability to comment starting at first column instead of first non whitespace character

aleffert:
I generally prefer my block comments to all start in the first column rather than before the actual content. This adds an option to support that. Also, my vim script is minimal so I have only the vaguest notion if this is correct. Thanks for the plugin!

### DIFF
--- a/doc/NERD_commenter.txt
+++ b/doc/NERD_commenter.txt
@@ -431,6 +431,9 @@ then the script would do a sexy comment on the last visual selection.
                                       around delimiters when commenting, and
                                       whether to remove them when
                                       uncommenting.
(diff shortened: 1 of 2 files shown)


nunoh:
+1 for this pull request. Would be greatly helpful for python commenting in order to avoid `E116: unexpected indentation (comment)`.

hackrole:
the pull request seems not merged, what's happend?

alerque:
Thanks for the contribution @alanfranz. As far as I can tell this is completely obsoleted by the code and config option merged from the much earlier PR #24. Since this does the same thing I'm going to close this PR, but please feel free to comment if there is something I'm missing.

PR #173, opened 2014-12-11
Fix Markdown escaping in README

jbodah:
The `<space>` command was not being properly escaped


alerque:
This was fixed in #122. Thanks for the contribution though!

PR #182, opened 2015-03-10
Use pep8 compatible comments for python

edgiese:
Python uses a lint program called pep8 that wants a space after the '#'.  Provide that by default.


edgiese:
There is already a pending pull request to do the same thing.

PR #183, opened 2015-03-17
according to pep8, we need a space after '#'

azdkj532:
according to python pep8 document, we need to insert a space after '#'.

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -303,7 +303,7 @@ let s:delimiterMap = {
     \ 'psf': { 'left': '#' },
     \ 'ptcap': { 'left': '#' },
     \ 'puppet': { 'left': '#' },
-    \ 'python': { 'left': '# ' },
+    \ 'python': { 'left': '# ', 'leftAlt': '#' },
     \ 'radiance': { 'left': '#' },
     \ 'ratpoison': { 'left': '#' },
     \ 'r': { 'left': '#' },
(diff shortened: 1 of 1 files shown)


davidjb:
+1  The documentation is at https://www.python.org/dev/peps/pep-0008/#block-comments:

> Each line of a block comment starts with a ``#`` and a single space (unless it is indented text inside the comment).

davidjb:
This looked good, but the issue with this change is that comments in Python may not necessarily start with a ``# ``.  If this change happens, then trying to toggle existing comments without a space will see them double-commented, and likewise, trying to uncomment will fail.  Changing the config to be:

    \ 'python': { 'left': '# ', leftAlt: '#' }, 

appears to do the trick.

Any comments with more than 1 space between the ``#`` and the text will have incorrect alignment if uncommented though. Regex isn't supported here, is it?

azdkj532:
@davidjb 
I had added it into the settings. :)

davidjb:
Great, thanks! +1 to this

hkupty:
how does this work with `g:NERDSpaceDelims=1`? doesn't this solve the problem without code change?

davidjb:
@hkupty, seemingly that'd require that option to be enabled specifically in Python files, which would require something to turn that on (eg ``autocmd FileType ...``).  Making the comment style fit PEP8 in the delimiter map from the start appears simplest & give the best performance too, presumably.

cmwslw:
Are there any reasons why this is not in master? Everything seems to work for me. I'm going to use this change for a while and I will post back if I find any issues.

davidjb:
No issues here; I've had the customisation in my vimrc since my comment
above.

scrooloose:
Merged. Apologies for the delay :)

PR #184, opened 2015-03-19
Update README.md

h0lyalg0rithm:
I spent about half an hour customising my mapping, just to realise,I had to restart vim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ See **|filetype-plugin-on|** for details, but basically, stick this in your vimr
 
     filetype plugin on
 
+Note: Close and reopen vim after installation, re-sourcing the vimrc file will not work.
+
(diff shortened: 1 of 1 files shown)


alerque:
Thank you for taking the time to contribute @h0lyalg0rithm. I think I'm going to call a pass on this. Even though I understand how this _can_ be an issue for some, it is also an assumed part of plugin management that you should follow the conventions for reloading them. In your case a restart fixed it, but some plugin managers have mechanisms for doing this. Rather than iterate through all of them or state the obvious (do what your plugin manager expects you to do) I think it's probably better in the interest of brevity to say nothing at all.

Feel free to comment back here if you still don't agree and I'll reconsider. Also if anybody else has feedback on this I'd love to hear it.

Issue #109, opened 2013-04-07
How does [count]<leader>c work?


trusktr:
When I try that, it just waits for my next keystroke or times out.

kennykaye:
:+1: This solves the problem, but still isn't displaying in README.

alerque:
This documentation issue was fixed in #209. Thanks to everybody that reported (and fixed) it!

PR #188, opened 2015-04-24
Fixed markdown display problem with angle brackets

marcodiiga:
Addresses https://github.com/scrooloose/nerdcommenter/issues/109


alerque:
Thanks for taking the time to contribute @marcodiiga. It seems this was previously fixed in PR #209, so I'm closing this as obsolete. If there are any further concerns please drop a comment.

PR #195, opened 2015-06-15
Changed python delimiter to matchpep8

gryftir:
per https://www.python.org/dev/peps/pep-0008/#id22 comments should be a # followed by a space.


georgexsh:
+1

jonafato:
#183 has been merged, so this change is now in `master`.

PR #189, opened 2015-04-29
edit python, add jinja

tlapicka:
pep8 rule requires space after hash

PR #196, opened 2015-06-23
add delimiters for Jinja

georgexsh:
ref: http://jinja.pocoo.org/docs/dev/templates/#comments


alerque:

**About this PR:** I have merged this PR as is. Please consider installing the plugin from `alerque/nerdcommenter` for this and (so far) over two dozen other updates.

alerque:

It turns out this PR is actually a duplicate of #189 (at least the part of it I cherry picked). That PR had a different comment syntax than you proposed here, with this being an alternate. Can you review and let me know if the current state of affairs is okay? If not please open a PR based against master with the correct syntax and some reasoning why and I'll give it a gander. Thanks for taking the time to contribute @georgexsh.

PR #200, opened 2015-08-09
Add support for elixir-lang

DavsX:
Support elixir-lang comments


alerque:
Thanks for taking the time to contribute @DavsX. This appears to be a duplicate of #108 which I have merged. Also just a small note for next time: whitespace cleanup is appreciated but usually best added to a separate commit. In this case it would have created a merge conflict as somebody else also did the same cleanup.

Let me know if there is anything else to fix!

PR #208, opened 2015-09-30
add elmlang support

rgscherf:


alerque:
Thanks for taking the time to contribute @rgscherf. I have cherry picked the main commit out of here. I did not include the extra space as there is a setting in this plugin for whether to include the space or not. This has come up in several other languages. If this is not satisfactory let's get an issue going about how to deal with it properly rather than having some one way and some another.

Issue #95, opened 2012-11-09
problem with vim  7.0.237


ralyodio:
```
Error detected while processing function NERDComment..<SNR>11_CommentLinesSexy..
<SNR>11_LeftMostIndx:
line   13:
E685: Internal error: get_tv_string_buf()
E116: Invalid arguments for function s:Left({'alt': 1}), s:Right({'alt': 1}), th
eLine))
E116: Invalid arguments for function s:IsCommented(s:Left({'alt': 1}), s:Right({
'alt': 1}), theLine))
E15: Invalid expression: a:countCommentedLines || (!s:IsCommented(s:Left(), s:Ri
ght(), theLine) && !s:IsCommented(s:Left({'alt': 1}), s:Right({'alt': 1}), theLi
ne))

```

alerque:
Are either of you still having this issue? I'm not seeing this now on much newer vim versions.

PR #221, opened 2015-12-28
Hack for redhat vim 7.0 with dict constants.

guobin2312:
Failing with E685: Internal error: get_tv_string_buf()

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -22,6 +22,13 @@ if v:version < 700
 endif
 let loaded_nerd_comments = 1
 
+" Section: dict constant init{{{2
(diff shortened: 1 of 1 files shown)


alerque:
Is this still an issue in current vim versions?

alerque:
Also is this related to #95?

guobin2312:
This is only a problem I found on old centos 5 vim 7.0. Looks very much
like #95.

I don't think it's worthwhile to include this hack: I put it here so
people can try it out.


On 05/24/2016 09:05 AM, Caleb Maclennan wrote:

Also is this related to #95<https://github.com/scrooloose/nerdcommenter/issues/95>?

PR #227, opened 2016-01-20
Add Rust and TypeScript rules

DenisKolodin:


alerque:
Thanks for taking the time to contribute @DenisKolodin. I've merged this, but there was a previous PR (#222) that had support for Rust and had the comment defaults the other way around. I kept that version, but if that is not satisfactory please raise the discussion here.

PR #228, opened 2016-01-21
Add comment syntax for SILE

alerque:
Sile is a typesetter somewhat similar to LaTeX that uses two different
possible input formats. Input can be XML which is already covered or
there is a TeX like input format. The normal file extension for the
latter is .sil.


alerque:

**About this PR:** This was the itch that I scratched myself ;-)

Issue #202, opened 2015-08-11
should merging of #183 necessary?


yuex:
Right, according to pep8, we need `# ` instead of `#`. But this kind of behavior can be implemented by `let g:NERDSpaceDelims=1`. Actually, isn't it necessary to add a space between the comment delimiter of any language to make it clean? The requirement of pep8 should not make python a special case. Or at least the implementation should take into account of `g:NERDSpaceDelims`. Someone argues that #183 makes it simpler for python commenting. But Before the merging I just need to config

    let g:NERDSpaceDelims = 1

now I need to config

    let g:NERDSpaceDelims = 1
    let g:NERDCustomDelimiters = {'python': {'left': '#'}}

besides two hours wasted to locate the reason why `<Leader>cc` only cause python comments two spaces instead of one.

cherrot:
Agree. 
Now I have to config for python specifically. It is annoying, really.

ryneeverett:
Is making `NERDSpaceDelims=1` the default a potential option? That would be pep8-compliant by default *and* avoid making python a special case. And as pointed out above:

> Actually, isn't it necessary to add a space between the comment delimiter of any language to make it clean?

ryneeverett:
On a somewhat tangential note, the current fix doesn't seem to even address pep8 block requirements:

> Block comments generally apply to some (or all) code that follows them, and are indented to the same level as that code. Each line of a block comment starts with a # and a single space **(unless it is indented text inside the comment)**.

Emphasis mine. The current implementation yields:

```py
# some_block = {
    # 'a': 1,
    # 'b': 2}
```

in which pep8/pycodestyle tool correctly identifies "unexpected indentation", rather than:

```py
# some_block = {
#     'a': 1,
#     'b': 2}
```

qalk:
I think you can meet Pep8 requirements with:

    let NERDDefaultAlign="left"

Put it inside .vim/after/ftplugin/python.vim to avoid changing the setting for other filetypes.

PR #230, opened 2016-03-02
updated ruby comment to fit with ruby style guide

sbdchd:
[Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide#hash-space) prefers a `#` with a space instead of a plain `#`.

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -316,7 +316,7 @@ let s:delimiterMap = {
     \ 'rib': { 'left': '#' },
     \ 'robots': { 'left': '#' },
     \ 'rspec': { 'left': '#' },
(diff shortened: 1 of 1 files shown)


alerque:
I have gone ahead and merged this as it seems to be the same  solution currently in place for Python and it's PEP8 style guide. That being said long term I think it's the wrong solution. Please see #202 for discussion on what the correct solution should be. Thanks for the contribution @sbdchd, and please do join with an opinion on how this common issue should be fixed more properly.

chdiza:
That ruby style guide has nothing at all like the sort of "authority" that PEP8 has.  Please revert this.

alerque:
@chdiza A later PR already replaced this and used the alt delimiters to add the Ruby block comment syntax. Please check the current version.

PR #234, opened 2016-04-15
Start Insertion in Append Mode When no Right Delim

alexander-alzate:


alerque:
Hey @alexander-alzate thanks for taking the time to contribute. I've recently been added as a maintainer on this project and am working through the backlog of contributions. Can you explain to me when this would be useful? In my testing I seem to already have the results you describe without having merged this commit, so I'm not quite sure what I'm testing for. In what scenario does this change the behavior?

alexander-alzate:
Hi @alerque,
When calling `<Leader>cA NERDComAppendCommen`, `startinsert` uses the actual position in order to edit, but when you have a right delimiter for single line comments (//), which is a <space> by default, you need to append text at the end of the line.

Without `startinsert!`, `|` represents the cursor position
~~~{.php}
$someVar; //| <- There is a trailing space
~~~

With `startinsert!`
~~~{.cpp}
std::String someChars; // |<- No trailing spaces
~~~

This is annoying when trying to avoid trailing spaces.

alerque:
Thanks @alexander-alzate. I'm actually away from any computer where I can test this for a few days but will have a look again next week based on that info.

alerque:
I got a second to review and it makes sense to me now. Thanks again for the contribution @alexander-alzate.

PR #237, opened 2016-05-23
Save people time

zwhitchcox:
Save people a command and keep them in their current directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,7 @@ The NERD Commenter requires Vim 7 or higher is comprised of 2 files:
 See **|add-local-help|** for more details. 
 
 ### Pathogen Install
-1. Navigate to your **[VIM_ROOT]/bundle** directory in your local Vim setup
-2. git clone https://github.com/scrooloose/nerdcommenter.git
+Paste this in your terminal: git clone https://github.com/scrooloose/nerdcommenter.git ~/.vim/bundle/nerdcommenter
 
 ### Post Install  
 After the **'Classic'** or **'Pathogen'** install, make sure that you have filetype plugins enabled, as the script makes use of 


alerque:
I'm going to call a pass on this. Sure it could save _some_ people time, but also introduces confusion. Since it's not a common `git clone` command  format it would be easy for somebody to bungle it up. The separate step of cd'ing to the right place both makes sure they know what they are doing and allows for installs which are not in the  path you hard coded (which by the way would break on Windows or for Neovim and in several other cases). Thanks for taking the time to contribute, but I don't think this improves the situation for everybody. Also anybody that understands this difference probably has a plugin manager going anyway.

zwhitchcox:
I have to, respectfully, disagree with you on this...I think it is a fairly common format for anyone even somewhat familiar with git. There is really no reason to change to the directory and is a pain for people who are frequently switching machines and are just getting started with the plugins, speaking from personal experience. I've never installed on Windows, but I think this would merit a case where you could have a Windows and a Unix version. It is a lot more convenient for me to just search for the repository on google, and copy and paste the plugin installer, and this impedes my workflow when doing so. Again, it's not like it's a huge deal, but it is kind of a pain.

I often only install plugins when I'm ready to use them to try and keep my machine clean, and to do this, I have only a few plugins and stuff that automatically get installed, and I only do this after several times of installing them and deciding they are necessary. Nerdtree is one that has made it into the automated installation process, but for a while there, I was copying and pasting every time, and I would have to figure out how to install it, which was a pain for a newish user, and this was just one of the problems that made it inconvenient.

Anyway, I was thinking of making my own plugin for navigation anyway, so I guess this can just be another reason to do that.

I have seen several other plugins do this, and they were much more convenient to install.

alerque:
Tell you what: I still disagree, but I don't have time to argue it right now. What I'll do for now is ① re-open this PR so that it hits the radar again and ② request that anybody reading this that understands the ins and outs of plugin install methods takes a minute to review and comment here.

If it turns out I'm the only one that sees this as a problem we'll go ahead with a merge. If other folks see the pitfalls and think the more verbose explanation is better we'll leave it the way it is. And if anybody has time to do a better/more through rewrite of the install instructions that would be helpful.

alerque:
Also, I merged a PR from somebody fixed a ton of spelling and other issues in the README and plugin docs so this now has merge conflicts and will need to be rebased.

zwhitchcox:
Cool! I went ahead and created a new [pull request](https://github.com/scrooloose/nerdcommenter/pull/241), taking the things you mentioned into account. I think it was a good call to make it an alternative, and I think it turned out much more explanatory and user friendly.

PR #100, opened 2013-01-09
Support for racket language.

pwightman:

PR #239, opened 2016-05-24
Reapply PR #183 for Python comments, PR #100 undid this

davidjb:
Pull request #100 for Racket undid the Python change, see the difference at https://github.com/scrooloose/nerdcommenter/commit/8d3dcc26a6c6b7cb98669bb6a6f9bdbc23f9e770.

This reapplies the patch from #183.

CC @alerque

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -320,7 +320,7 @@ let s:delimiterMap = {
     \ 'psf': { 'left': '#' },
     \ 'ptcap': { 'left': '#' },
     \ 'puppet': { 'left': '#' },
-    \ 'python': { 'left': '#' },
+    \ 'python': { 'left': '# ', 'leftAlt': '#' },
     \ 'racket': { 'left': ';' },
     \ 'radiance': { 'left': '#' },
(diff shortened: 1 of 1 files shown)


alerque:
Thanks @davidjb. This appears to be a merge conflict resolution error on my part. Actually there is a bigger problem here in that there are people pushing both ways for including spaces in delimiters vs. not, but that should be worked out and done officially rather than introducing the change by accident in a merge. I've merged this back in for now, thanks for the catch.

davidjb:
Cheers @alerque.  I suppose it depends on whether given Python code is following PEP8 or not.  Since that's only a style guide (widely adhered to, though), there's no guarantee that a given file is following it.  It's easy enough to override nerdcommenter's style though if someone wanted the default to be without a space.

alerque:
It's slightly more complicated that than because an argument could be made that this plugin should recognize comments the way the compiler does and only optionally also follow style guides or user preferences. As it stands _uncommenting_ some valid comment is kind of a kludge.

Lets take the discussion about how this _should_ be implemented to #202. I don't want to fiddle with this until there is some consensus about the best way to actually solve this rather than playing tug-of-war with the plugin defaults.

PR #241, opened 2016-05-24
alternative pathogen download instructions

zwhitchcox:
Made the Unix pathogen download instructions alternative, rather than the only way

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,13 +12,13 @@ The NERD Commenter requires Vim 7 or higher is comprised of 2 files:
 ###### Unix
 
 ```sh
-git clone https://github.com/scrooloose/nerdcommenter.git ~/.vim/bundle
+git clone https://github.com/scrooloose/nerdcommenter.git ~/.vim/bundle/nerdcommenter
 ```
 
 ###### Neovim
 
 ```sh
-git clone https://github.com/scrooloose/nerdcommenter.git ~/.config/nvim/bundle
+git clone https://github.com/scrooloose/nerdcommenter.git ~/.config/nvim/bundle/nerdcommenter
 ```
 
 #### Vim-Plug Install
@@ -48,8 +48,7 @@ See **|filetype-plugin-on|** for details, but basically, stick this in your vimr
 ###### Unix
 
 ```sh
-curl -fLo ~/.vim/plugin/NERD_commenter.vim --create-dirs \
-    https://raw.githubusercontent.com/scrooloose/nerdcommenter/master/plugin/NERD_commenter.vim
+git clone https://github.com/scrooloose/nerdcommenter.git ~/.vim/plugin/nerdcommenter
 ```
 
 ###### Windows (PowerShell)
@@ -63,8 +62,7 @@ $uri = 'https://raw.githubusercontent.com/scrooloose/nerdcommenter/master/plugin
 #### Neovim
 
 ```sh
-curl -fLo ~/.config/nvim/plugin/NERD_commenter.vim --create-dirs \
-    https://raw.githubusercontent.com/scrooloose/nerdcommenter/master/plugin/NERD_commenter.vim
+git clone https://github.com/scrooloose/nerdcommenter.git ~/.config/nvim/plugin/nerdcommenter
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -1,35 +1,72 @@
 # Installation  
 
-_\*For simplicity, let's use **[VIM_ROOT]** which will mean ~/.vim (\*nix) or ~/vimfiles (Windows)_
-
 The NERD Commenter requires Vim 7 or higher is comprised of 2 files:  
 
     plugin/NERD_commenter.vim  
     doc/NERD_commenter.txt 
 
-### Classic Install
+## Via Plugin Manager (Recommended)
 
-1. Extract the plugin files into your **[VIM_ROOT]**/plugin and **[VIM_ROOT]**/doc folders respectively.
-2. Finish the install by updating your help files.  Run:
+#### Pathogen Install
 
-<pre><code>:helptags [VIM_ROOT]/doc</code></pre>
+###### Unix
 
-See **|add-local-help|** for more details. 
+```sh
+git clone https://github.com/scrooloose/nerdcommenter.git ~/.vim/bundle
+```
 
-### Pathogen Install
-1. Navigate to your **[VIM_ROOT]/bundle** directory in your local Vim setup
-2. `git clone https://github.com/scrooloose/nerdcommenter.git`
+###### Neovim
 
-### Vundle Install
+```sh
+git clone https://github.com/scrooloose/nerdcommenter.git ~/.config/nvim/bundle
+```
+
+#### Vim-Plug Install
+1. Add `Plug 'Plug scrooloose/nerdcommenter'` to your .vimrc file.
+2. Reload .vimrc (`:so %` from vimrc, or just restart vim)
+3. Run `:PlugInstall`
+
+#### Vundle Install
 1. Add `Plugin 'scrooloose/nerdcommenter'` to your .vimrc file.
 2. Run `:BundleInstall`
 
-### Post Install  
+#### NeoBundle Install
+1. Add `NeoBundle 'scrooloose/nerdcommenter'` to your .vimrc file.
+2. Run `:NeoUpdate``
+
+#### Post Install  
 After the **'Classic'** or **'Pathogen'** install, make sure that you have filetype plugins enabled, as the script makes use of 
 **|'commentstring'|** where possible (which is usually set in a filetype plugin). 
 See **|filetype-plugin-on|** for details, but basically, stick this in your vimrc:
 
     filetype plugin on
+
+## Install Manually
+
+#### Vim
+
+###### Unix
+
+```sh
+curl -fLo ~/.vim/plugin/NERD_commenter.vim --create-dirs \
+    https://raw.githubusercontent.com/scrooloose/nerdcommenter/master/plugin/NERD_commenter.vim
+```
+
+###### Windows (PowerShell)
+
+```powershell
+md ~\vimfiles\plugin
+$uri = 'https://raw.githubusercontent.com/scrooloose/nerdcommenter/master/plugin/NERD_commenter.vim'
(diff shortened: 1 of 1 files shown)


alerque:
Hey this is better that #237, but still has at least one issue. As I commented before, even on Unix the VIM_ROOT is not always going to be `~/.vim`. Mine for example is `~/.config/nvim`. Either we need to stick to the `[VIM_ROOT]` conversion in all the examples or ditch it entirely, but I don't like the mix and match style where some commands use it and others hard code a value that won't be right for everybody.

How about including a line to actually set a  VIM_ROOT variable (with tips on how to adjust for any platform)? Then an full terminal command could be given that uses it that would be copy/pasteable.

zwhitchcox:
Ok, so I made the changes you suggested and amended them, so it will all be as one commit.

Just to explain my thinking a little: I think people who have changed their vim root are probably aware of that and can make adjustments for that. My concern is for people just getting started and might not be able to figure out what their vim root is or the right shell commands to navigate there. So, to accommodate everyone and still make it where new users can just copy and paste. I think 99% of people will be able to the copy and paste, and the 1% will be able to figure it out, especially with the comments.

zwhitchcox:
This is all me speaking as a completely self-taught programmer without mentors or anything. I've had to go in and figure out about a million of these things, and this is, personally, what I would want

alerque:
> I think people who have changed their vim root are probably aware of that and can make adjustments for that.

Of course I agree that the people that change this themselves would be able to fix it themselves. The thing is some versions of vim (e.g. Neovim) ship out of the box with different locations and some distros change the location for plain vanilla vim in their packaging process, so  some users with no clue which way is up will not have the location you think of as the default.

Thanks for the updates, I'll review in a minute.

zwhitchcox:
Ah, cool, I hadn't heard of Neovim. And those are some good points. That complicates things a bit. Seems pretty dumb of those distros, but I think it would be safe to say the root of the "standard" installation of vim is in `~/.vim`. Still, it would be nice to automate the process somehow, if we could dynamically find the vim root.

alerque:
Automating the process is what most plugin managers do. They make it really easy to add, remove, and keep fresh a host of plugins. My favorite is vim-plug if you're looking for a place to start.

zwhitchcox:
Ah, good idea. I'll look through their code, and we can just steal how they do it ;)

zwhitchcox:
Interesting, for [their installation method](https://github.com/junegunn/vim-plug/blob/master/README.md#installation), they give you three methods, one for regular vim, one for neovim, and one for Windows Powershell. 

I think if we offered all of those options, that would be enough, and if there are other corner cases, we can address those as they come.

alerque:
The breakdown for their own install makes some sense. That might be a better way to organize the install instructions. I suggest it should also be arranged so that the structure is something like:

  * Install via plugin manager (recommended)
      * vim-plug
      * vundle, neobundle, etc.
      * pathogen
      * etc.
  * Install manually
      * vim
          * Unix
          * Windows
      * neovim

Does that sound reasonable? Do you want to work on that or do you want me to merge this as a starter point and hack on it myself?

alerque:
That's looking better and better. I don't think the autoload folder is the right place to put this though. Remember the vim-plug install in a bootstrap procedure after which in manages itself, which is not the case for us. We still want users to do the git clone of the repo and put it in the plugins folder, otherwise they won't get the docs etc.

zwhitchcox:
Ok, I think I implemented all the changes. Let me know what you think

alerque:
Sorry I missed your 7a9edfe commit. I pulled it at 0cbbcb5, merged and ran with it from there. That being said I think I corrected that mistake and a few others. Let me know what you think of the current form. I know you don't like the two step CD thing, but even the author of pathogen recommends it and it has the  huge advantage of throwing an error if the user is in the wrong place. Git sometimes makes the parent directories even if they are all wrong and the potential confusion is too great. You can copy and paste the path onto the clone line if the extra step is too much trouble. Or if cd's bother you, wrap it in a `()` subshell or set your shell to default to using `pushd` so you can `popd`. There are lots of ways to fix that if you know what you are doing, but I think the the least likely to break solution needs to be there for the confused new users.

zwhitchcox:
I don't quite understand what you mean by "if the user is in the wrong place". Isn't the whole point that it doesn't matter where the user is, because it will always copy it to the right directory? And git will make the parent directories even if they are wrong? Aren't we specifying the directory as the second argument to get clone? If we were to use the platform-specific instructions, I think it would fix all those concerns. The only disadvantage I can see would be the added length to the instructions, which I don't really like, personally, but I think it could easily be ignored by the user not using pathogen.

As it stands, the pathogen instructions still don't account for the situation where a user is using windows or neovim, so I think they're flawed either way. Changing directories will have no impact on this, except to perhaps catch the case where a user is using Neovim and discovers the directory doesn't exist. Personally, I think it still would be prudent to add the platform-specific instructions for pathogen and just make it a one-liner. It would catch more errors and give everyone a one-liner they can just copy and paste into their terminals, but that's just me.

Everything else looks great to me, and I, personally, would just use the manual installation, because I think that's the easiest and doesn't require any dependencies, so it doesn't really matter to me whether the pathogen instructions are even there, really.

zwhitchcox:
Honestly, I don't even understand why pathogen exists if you can just clone files into plugin. I probably will stop using it. I did not know you could do that. And after seeing that, I would probably just take it out altogether, actually.

alerque:
We _can't_ just clone into plugin because vim doesn't look inside subfolders there. Pathogen exists no manage the paths that vim looks in so that you _can_ have a folder that you clone to and have side by side dirs for each plugin. That's what it does that the manual install can't. Notice I replaced the manual install with curl and two paths. Honestly I'd just assume drop the manual install directions: nobody actually wants to do that and if they do they can be on their own.

Git cloning to a specific path can potentially clone so a directory too high or two low. For example if the target folder doesn't exist, it will create it and use that as the last segment.

```sh
# Normally this would create a subfolder inside bundle for the clone
git clone <repo_url> ~/.vim/bundle/

# But what happens if ~/.vim/bundle doesn't exist?
# Oh look now you have the repo checked out in ~/.vim/bundle itself! Add the target you say?
git clone <repo_url> ~/.vim/buldle/<repo_name>

# Sure. But what if <repo_name> happened to exist?
# Now you have ~/.vim/bundle/<repo_name>/<repo_name>
```
Do you see where that is going? There just isn't a right answer for what to put there that works for everybody. Demonstrating it with `cd` gives people a chance to understand what they are doing and get a sensible error if it goes wrong.

zwhitchcox:
Well, I just cloned the project into my plugin directory and it worked, so maybe that's a new feature in vim. Anyway, I'll always be using the latest vim, so for me, personally, it doesn't make a difference.

And if you do your second option, and the repo does exist, git throws an error for me. I don't know maybe that's a new feature too.

zwhitchcox:
Basically, git is going to use the path you give it. So, `git clone http://path.com/repo.git ~/.vim/bundle` would try to clone the repository as the bundle directory. If the bundle directory already exist, it would throw an error, which it just did for me.

zwhitchcox:
And I don't understand why you would say no one wants to do the manual install. That seems like the easiest option to me. It literally requires you to copy and paste one line of code without any dependencies whatsoever.

alerque:
> Well, I just cloned the project into my plugin directory and it worked

I'd be willing to bet it didn't. Recent versions of vim do find `*.vim` files in subfolders of `plugin`, but they don't load the whole package they way they should. For example if you take a new unix user account (i.e. with no other vim setup) and clone into `~/.vim/plugin/nerdcommenter`, at least in 7.4 you'll end up with some basic commenting functionality (minus file type detection). You will not, however, end up with the documentation being  available. `:help nerdcommenter` give you an error.

What's going on there in that the run time path isn't getting set right. See `:set rtp` for a list of places it's looking, in each of which it expects to find a certain structure. None of those locations makes it possible to just throw in a directory per plugin and have everything work. This is what Pathogen takes care of while letting you manage your own folders or git clones. Other plugin manages also take over managing `rtp` but also manage they directory structure and handle updating.

> if you do your second option, and the repo does exist, git throws an error for me

This behavior changed—about 1.7 if I remember right.

> Basically, git is going to use the path you give it.

No not always. And when it does, it can still be wrong.

Here's another case. The user could have a vim compiled to look in `~/.config/vim/bundle/nerdcommenter`, but we tell them to clone for Pathogen into `~/.vim/bundle/nerdcommenter`.  Their clone will work, but their plugin will not. Another possibility is they typo it. Again the clone works including making parent dirs. No indication is given suggesting they've strayed. Now they are send in circles trying to find their mistake. Much easier to start them out `cd`ing to where they should be so they figure out up front if they have what they need or whether they need to be in a different place.

> And I don't understand why you would say no one wants to do the manual install.

Because it's hard to figure out all the details, brittle when you're done, and you quickly end up out of date. I have 67 plugins right now. Who wants to track down the file listings and URLs to make sure they are all up to date and copied into the right places.

If you can figure it out and want to do it manually, knock yourself out. But for our new users I think we need the most failsafe thing, not necessary the thing with the least commands. If easy is the goal, plugin managers do this well. The manual instructions are most likely to be followed by people just getting started with vim addons that haven't been convinced they need a plugin manager yet. For them not breaking things is more important that tricksy one-liners. Anybody else is going to know what they want and do whatever they want anyway.

zwhitchcox:
Ok, so why not just make it 

`$(cd ~/.vim/bundle && git clone https://github.com/scrooloose/nerdcommenter.git)`

That seems to solve every problem, right?

alerque:
You wouldn't need the `$`, that is command substitution. You can use just the subshell (`(`). But I still object on the gounds that most people don't know what's going on there and are likely to bungle it. Making people copy and paste something they don't understand means that when it doesn't go right (say their vim is in a different place) then they don't understand how to fix it.

Folks like you that can hack around in a shell are free to add the extra parens and ampersands to make it run in a single copy and paste if you like. Start a list of all the one-liners you use to setup a new machine. Or you know, use a plugin manager. But I don't think it's worth the risk of confusing everybody else by doing something fancy pants that they don't expect or know how to fix.  All this discussios has only served to solidify the idea it my head that the most obvious commands (even if it's two commands instead of one) is the best way to handle this. Baring input from other parties showing this is a point of disagreement, this is the way it's going to be for now. Sorry.

PR #248, opened 2016-05-26
Fix rust Alt comments

ARM9:
/// are single line documentation comments.

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -348,7 +348,7 @@ let s:delimiterMap = {
     \ 'robots': { 'left': '#' },
     \ 'rspec': { 'left': '#' },
     \ 'ruby': { 'left': '#', 'leftAlt': '=begin', 'rightAlt': '=end' },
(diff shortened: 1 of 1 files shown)


alerque:
Hey @ARM9 thanks for taking the time to contribute. You're probably right here and I'm no expert on Rust, but before I blow it away can you give me a brief explanation here? The previous contributor that set this entry up linked to [this page in the docs](https://doc.rust-lang.org/book/comments.html) that says nothing about `/* */` style block comments. What's going on here?

alerque:
Hey @Ilphrin, any comment on this?

ARM9:
I see, that page seems to omit multiline comments for some reason (perhaps they don't update it regularly). See https://doc.rust-lang.org/reference.html#comments

alerque:
I can buy that. @llphrin and anybody else that prefers the docstring comment format as the alternate, you can set it up in your vimrc file like this:

```vim
let g:NERDCustomDelimiters = { 'rust': { 'left': '//', 'altLeft': '///' }
```

Ilphrin:
Great! :) I didn't know if there is a way to manage all kind of Rust's comments, but the common syntax is cool, thanks!

PR #249, opened 2016-05-31
Add hook functions for NERDComment and SwitchToAlternativeDelimiters

xream:
This is useful when editing files like [vim-vue](https://github.com/xream/vim-vue).

An example: [file-ini2-vim](https://gist.github.com/xream/c474a1adffeb6f70daa6a7ddc22386e0#file-init2-vim).

In hook functions, we can change the filetype, comment and change it back.

I could not come up with any better solution. If there's any better solution, please let me know.

Thank you for this wonderful plugin.

### DIFF
--- a/doc/NERD_commenter.txt
+++ b/doc/NERD_commenter.txt
@@ -28,8 +28,9 @@ CONTENTS                                               *NERDCommenterContents*
             3.2.11 Use alternate delims map...|NERDComAltDelim|
             3.2.12 Comment aligned maps.......|NERDComAlignedComment|
             3.2.13 Uncomment line map.........|NERDComUncommentLine|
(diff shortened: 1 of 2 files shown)


alerque:
These look great @xream, and thanks for taking the time to contribute. Eventually this will need a bit of an overhaul, but I think it's a step in the right direction for handling different language blocks embedded in the same file.

Could I ask that you also document this feature in the the vim help docs? If this is going to really be useful to others it needs to be discoverable and at least listing the available hooks, where they run, and mentioning  what they might be good for will make that possible. You should be able to edit the text file in the repo and add a commit to this PR by pushing to the same branch.

xream:
Done 😄. I'm glad to contribute. I am not a native English speaker. There may be some English grammar errors 😅.

Issue #194, opened 2015-06-05
Do not consider blank lines uncommented when toggling


DelvarWorld:
If you make a visual selection of a **commented** block of text, **then include an empty line above the selection** and hit `<leader>c<space>` to toggle comments, NERDCommenter appears to think the blank line is uncommented, and double-comments the selected block.

It would be much nicer if the algorithm that determines if something is commented (which appears to use the first line of a visual selection) read lines until one of them wasn't blank.

alerque:
This actually is a bit of a catch 22. It works as expected for _multi-line comments_. If you use a multi-line syntax, including a blank line above a selection does the right thing already. It's only when using line-wise delimiters in leu of block delimiters that this seems to be a problem. I'm not sure how the logic for this would work given that a line-wise check for comments wouldn't find one (which is correct).

alerque:
@DelvarWorld Somebody has stepped up and implemented something like this. Would you be able to checkout the code from Pull Request #250 and see it it works for your use case and offer any feedback you  make have?

alerque:
This is not an option you can set if you want this behavior. Thanks to @alexconst for the implementation.

PR #250, opened 2016-06-06
Allow commenting empty lines given configuration

alexconst:
If `g:NERDCommentEmptyLines=1` then it will comment empty line, which is useful when commenting blocks.
When uncommenting it will delete any trailing whitespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,9 @@ let g:NERDAltDelims_java = 1
 
 " Add your own custom formats or override the defaults
 let g:NERDCustomDelimiters = { 'c': { 'left': '/**','right': '*/' } }
+
+" Allow commenting and inverting empty lines (useful when commenting a region)
+g:NERDCommentEmptyLines = 1
 ```
 
 ### Default mappings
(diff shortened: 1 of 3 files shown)


alerque:
Hey @alexconst thanks for taking the time to contribute. I'm glad somebody is taking up this feature as it was requested some time ago (see #194). I think this will basically address that concern.

However there are a couple things I would request before merging this.

1. Can you split the whitespace function out into a different commit and PR? I have some questions about how that might impact some languages and I think we need to poll for community feedback separately from this PR (which I think I can push through right away).

2. Thanks for documenting this in the readme, but the real place in needs to be documented is the inline help docs. See the text file in the docs folder of the repo. Can you document this in there so the help system will be properly documented?

3. I think the default value for this should probably be explicitly set. This just seems like good  practice and would be make the whole thing more transparent to people trying to read through the code.

alexconst:
1. Done. And to simplify things I made it as an option that is disabled by default. This way it can be merged without breaking anything. https://github.com/scrooloose/nerdcommenter/pull/251

2. Done.

3. I don't understand what you mean. How do you want it to be explicitly set? In `NERD_commenter.vim` the option is set to `0` if not explicitly set by the user. This seems reasonable to me.

Thanks

alerque:
Awesome thanks. And about 3, that was my bad I missed the code where that was getting initialized to a default value. Sorry about that.

PR #251, opened 2016-06-08
Add option to trim trailing whitespace

alexconst:
Add option g:NERDTrimTrailingWhitespace to allow deleting trailing
whitespace when uncommenting a line.


alerque:
Thanks for splitting this out. Unfortunately since the code and docs are touching about the same lines in the source this created a merge conflict when I merged the other code. If you have a chance to rebase it on master and mark how the conflict should be sorted that would be great. If not I'll be able to do it, but it might wait a while as I'm on the road.

alexconst:
@alerque I think it should be ok now. If not then let me know.
Thanks

PR #254, opened 2016-06-28
README.md: fix vim-plug installation instructions

kalbasit:


alerque:
Thanks for the fix! Sorry I've been traveling for several weeks so am seeing this late, but the fix is applied now.

PR #255, opened 2016-07-06
Add hive filetype

ffabulous:
HQL(hive query language) syntax is almost equivalent to SQL.


alerque:
Thanks for the contribution. Sorry I was traveling so didn't get to this for a while, but it's in now.

PR #265, opened 2016-09-09
added python multiline commenting

KingEdwardI:


KingEdwardI:
able to comment, but not able to uncomment python.

alerque:
Hey @KingEdwardI are you sure you were using the correct configuration and command combination for commenting from inside a multi-line comment in general?

KingEdwardI:
yes, the configuration and command combo are correct. multiline comments in python leave a single quote at the beginning of the line when uncommenting, behavior in javascript is similar, but does not work in python. 
Javascript Commented: 
screen shot 2016-09-11 at 3 14 52 pm
Javascript Uncommented:
screen shot 2016-09-11 at 3 15 09 pm
Python Commented: 
screen shot 2016-09-11 at 3 16 44 pm
Python Uncommented:
screen shot 2016-09-11 at 3 17 01 pm

PR #266, opened 2016-09-09
add support for turtle (.ttl) files

Aluriak:
Turtle format is defined by W3C, and is one of the well-known formats for RDF encoding.

As in Python, comments are simply `#`, and following space is not mandatory, but is to me a readability gain.


alerque:
I'd be happy to add the support for support for Turtle, but can you please drop the second commit with the space here before I merge this? All of the comment delimiters do not include spaces. Those who prefer them for the readability (myself for one) can set this using a configuration option:

```vim
let g:NERDSpaceDelims = 1
```

Aluriak:
Well, that's done !

Issue #268, opened 2016-09-22
Additional trailing whitespace after uncommenting block


chipironcin:
Use this block of code:

```
configureOptions() {
  console.log("Hello World");
  return list;
}
```

Select the two lines inside the function in Visual Mode.
Comment them in a block with `<leader>cs`
Select the two lines inside the function in Visual Mode.
Uncomment them in a block with `<leader>c<space>`

You will see a trailing whitespace that appeared at the end of the last line that was commented/uncommented.

This should be avoided, if anyone can find a workaround or can spot the bug it would be awesome.

chipironcin:
I have enabled the following options in my vimrc
```
" Add spaces after comment delimiters by default
let g:NERDSpaceDelims = 1
" Use compact syntax for prettified multi-line comments
let g:NERDCompactSexyComs = 1
" Align line-wise comment delimiters flush left instead of following code indentation
let g:NERDDefaultAlign = 'left'
" Set a language to use its alternate delimiters by default
let g:NERDAltDelims_java = 1
" Allow commenting and inverting empty lines (useful when commenting a region)
let g:NERDCommentEmptyLines = 1
" Enable trimming of trailing whitespace when uncommenting
let g:NERDTrimTrailingWhitespace = 1
```

chipironcin:
Solved, thanks!

PR #269, opened 2016-09-22
Remove trailing whitespace for top and bottom line

chipironcin:
Remove trailing whitespace for top and bottom line when uncommenting a sexy commented block.
Fixes https://github.com/scrooloose/nerdcommenter/issues/268

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -1615,6 +1615,16 @@ function s:UncommentLinesSexy(topline, bottomline)
         let theLine = s:SwapOuterPlaceHoldersForMultiPartDelims(theLine)
         call setline(bottomline, theLine)
     endif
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for the contribution. I'll review this for real next week when I'm not on the road. In the mean time have you checked whether it plays nice with the various user facing settings for how to handle spaces? How about Python and the current snafu going on in this plugin with the PEP8 comment standard?

chipironcin:
I have not checked anything about that, definitely will do.
Anyways this change is not that disruptive, I am just removing the trailing whitespace not only for the inner lines in the block but for the first/last lines too.

alerque:
Sorry this took me so long to get back to, but it's merged now. Thanks again for the contribution @chipironcin.

PR #271, opened 2016-10-05
add support for M4 macro processor

rohieb:
M4 knows about two types of ignoring text.  But the concept to which the GNU M4 manual refers to as ["comments"][0] (i.e., lines prefixed with hash signs) is actually telling the interpreter "ignore this text, but pass it to the output".  Most of the time, people don't want the macro comments in their output, so instead the [`dnl` ("discard to next line") macro][1] is usually used for comments in the document-my-thoughts sense.

[0]: https://www.gnu.org/software/m4/manual/m4.html#Comments
[1]: https://www.gnu.org/software/m4/manual/m4.html#Dnl

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -255,6 +255,7 @@ let s:delimiterMap = {
     \ 'lua': { 'left': '--', 'leftAlt': '--[[', 'rightAlt': ']]' },
     \ 'lynx': { 'left': '#' },
     \ 'lytex': { 'left': '%' },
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for this contribution. I'm going to go ahead and merge it now, but there is actually a problem that needs to be fixed in that this isn't the correct way to handle comments that require spaces as part of their syntax. Given that the whole plugin is broken the same way right now a fix for that doesn't need to hold this up, but just be aware the way this is handled will change eventually.

PR #275, opened 2016-11-15
Add support for molpro and vasp chemistry packages

alejandrogallo:
Molpro and vasp are electronic structure program
used in quantum chemistry.

I added support for input files of these programs.
The filetypes can be found in vim packages such as

https://github.com/alejandrogallo/vasp.vim

or

https://github.com/antoniogsof/vim-molpro


alerque:
Thanks for the contribution @alejandrogallo! The only thing I would ask is that you stick these in in alphabetical order to be consistent with the rest of the list. Can you rebase this commit to fix that or would you prefer I do it locally? (Easy for me to do and the commit will still have your name on the contribution but this PR won't show as being accepted.)

alejandrogallo:
Hi,

I am sorry for the inconvenience.
I would prefer that you do it locally if it's ok for you,
I don't care about my name, I just want the feature ;)

Thanks!

alerque:
Done.

alejandrogallo:
Nice thanks!

PR #279, opened 2017-01-05
Added filetype 'renpy' support.

komejikoishi:
Renpy is an extension of python syntax and it uses '#' for comments.


alerque:
This should not specify `leftAlt` since it is not enabling a different comment syntax.

With or without space is controlled by a different setting in this plugin, and the "with space" comment syntax is only forced for languages that either require it as part of the language spec or (looking at you Python) that bludgeon you into it through enforced style guides.

A solution for disabling the latter should be part of a different PR, see #202 and other issues for discussion.

PR #281, opened 2017-01-17
space for comments in elixir

techgaun:
I think `# ` is better than `#` and that's what I see in most Elixir codes and I think is more preferable https://github.com/elixir-ecto/ecto/blob/master/config/config.exs#L3

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -158,7 +158,7 @@ let s:delimiterMap = {
     \ 'eclass': { 'left': '#' },
     \ 'eiffel': { 'left': '--' },
     \ 'elf': { 'left': "'" },
-    \ 'elixir': { 'left': '#' },
+    \ 'elixir': { 'left': '# ', 'leftAlt': '#' },
     \ 'elm': { 'left': '--' },
     \ 'elmfilt': { 'left': '#' },
     \ 'ember-script': { 'left': '#' },


alerque:
With the exception of languages that _require_ the space to pass syntax validation, the space after comment markers is optional. As such the comment markers in this plugin use just the minimum syntax and use an  option system to allow users to add spaces if they wish. I personally prefer it to, but the correct way to do this is to set the preference in your vim rc file:

```vim
let g:NERDSpaceDelims = 1
```

Please see the [settings](https://github.com/scrooloose/nerdcommenter#settings) section in the README for more commonly used options and a way to override the default syntax modes for specific  languages of your choice.

Thanks for the contribution, but an option for this already exists and merging this would conflict with the way spaces are already handled per user preferences.

techgaun:
@alerque fair enough. I did see that section and I'll go with that preference then. My original assumption was that it was acceptable to have this as default based on https://github.com/scrooloose/nerdcommenter/blob/master/plugin/NERD_commenter.vim#L343 . I understand what you're saying and why that's preference there for python. I see space after pound as becoming more and more common on Elixir and most people prefer that. Also, I think some of the linters for Elixir complain about that (I know dogma does that). So I was hoping this could be default. Anyway thanks I can still use the setting you talked about. thanks!

alerque:
@techgaun It's there for Python and derivatives because in spite of it's name PEP8 is more than a style guide and a lot of stuff breaks if you don't follow it.

The way the plugin works right now putting a space as part of the delimiter doesn't just make it the "default", it makes it impossible to use any other way. Without the space in the delimiter anybody who wants it can turn it on. With it, even those who don't want it for some reason are stuck with it. That's not what we're here for.

There has been discussion before about how to fix this and the best concept I've seen would be to have the files array include information about style guides / most common customs for each language. However somebody would have to implement this. In the mean time just hacking more file types to force it on is not the way to go. 

Another stand by might be to turn that space preference on by default and let people that don't want it turn it off. However with thousands of people using this I wouldn't reverse the defaults lightly and we need to make sure that's the best solution.

Remember I'm in the spaces are better for almost all languages camp, so I'm not just being a grump because of my preferences. I'm just saying we can't rock the boat except carefully and just hard coding some file types doesn't give users the flexibility to make it do what they want (which is why some of us use vim in the first place).

techgaun:
@alerque I get your point. Thanks for the elaboration.

PR #283, opened 2017-01-19
Add multi-line comments for Elm

zack:
Technically the minimum necessary syntax for block comments requires
only a single hyphen (`-}`, `{-`), but I have found that using a double
hyphen plays a bit more nicely with the elmcast/elm-vim auto-formatting.
Further, uncommenting when using the single dashes leaves an extra level
of indentation (again, when using elmcast/elm-vim).

-- These are PR (not commit) comments:
Vim auto-removed extra whitespace. Let me know and I can remove those line
changes.

Here's an example of this multi-line comment in action and the resulting
auto-format from elmcast/elm-vim:

```
main : Program Never Model Msg
main =
    Html.program
        { init = ( initialModel, Cmd.none )
        , view = view
        , update = update
        , subscriptions = (\model -> Sub.none)
        }
```
becomes
```
{--
  - main : Program Never Model Msg
  - main =
  -     Html.program
  -         { init = ( initialModel, Cmd.none )
  -         , view = view
  -         , update = update
  -         , subscriptions = (\model -> Sub.none)
  -         }
  --}
```
versus using a single dash:
```
{-
   - main : Program Never Model Msg
   - main =
   -     Html.program
   -         { init = ( initialModel, Cmd.none )
   -         , view = view
   -         , update = update
   -         , subscriptions = (\model -> Sub.none)
   -         }
-}
```

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -159,7 +159,7 @@ let s:delimiterMap = {
     \ 'eiffel': { 'left': '--' },
     \ 'elf': { 'left': "'" },
     \ 'elixir': { 'left': '#' },
-    \ 'elm': { 'left': '--' },
+    \ 'elm': { 'left': '--', 'leftAlt': '{--', 'rightAlt': '--}' },
     \ 'elmfilt': { 'left': '#' },
     \ 'ember-script': { 'left': '#' },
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for the contribution @zack. Sorry it took me so long to review this, but it looks fine.

PR #284, opened 2017-01-21
Add gitignore support

alejandrogallo:
There is gitcommit but not yet a gitignore comment symbol.  I think it may come
in handy for those that define the gitignore filetype.

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -192,6 +192,7 @@ let s:delimiterMap = {
     \ 'gentoo-package-use': { 'left': '#' },
     \ 'gitcommit': { 'left': '#' },
     \ 'gitconfig': { 'left': ';' },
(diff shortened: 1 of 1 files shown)


plugin/NERD_commenter.vim
```diff
@@ -174,6 +174,10 @@ let s:delimiterMap = {
     \ 'focexec': { 'left': '-*' },
     \ 'form': { 'left': '*' },
     \ 'fortran': { 'left': '!' },
+    \ 'molpro': { 'left': '!' },
```
alerque: These look fine, but can you please place them in the list in alphabetic order?

alejandrogallo:
Done, sorry I messed up, only the gitignore commit was intended to be in
the pull-request, the other files where already joined in commit
e1af5da.

alerque:
Thanks @alejandrogallo that makes more sense now.

PR #302, opened 2017-06-06
Upward Commenting

Skian12:
When we use NERDCommenter, in basically [count] determines how many lines selected 'downward' from the cursor.

so, we have a question "Can we make [count] determine how many lines selected 'upward' from the cursor?"

and this pull request is realize the question.

If you want to use this feature, you just type one more <leader> key.

for example, if you want to use |NERDComComment| 'downward', then just type [count]<leader>cc. but if you want to use |NERDComComment| 'upward', then type [count]<leader><leader>cc.


Skian12:
sorry, this is not the version of the upward commenting

PR #307, opened 2017-08-19
Fix #278 Double space delimiter for some languages

### LINKED ISSUE #278: Double space delimiter for some languages with g:NERDSpaceDelims=1
Python and a few others with a space in the comment string end up having two spaces when setting
`let g:NERDSpaceDelims=1`

I suggest that g:NERDSpaceDelims=1 doesn't add an extra space when there already is one, that the interpretation of this option is to ensure that there is a trailing space, rather than adding one.

This allows for having a default trailing space for languages like Python where style guides recommend having a space without introducing double spacing for those who prefer to always have a trailing space which is what g:NERDSpaceDelims accommodates

Those who want double spaces could change the comment string to have two spaces, and these would need to adjust their settings, the rest wouldn't.

Obirvalger:


alerque:
This doesn't actually fix the problem (as discussed in #202), either permanently in dealing with languages where the  space is part of the syntax or temporarily (in the way suggested in issue #278). This is just a hack that can already be added to a user's RC file and (as I discovered when I tried something similar a while back) will break this plugin for lots of users.

I appreciate the contribution @Obirvalger, but this issue needs a proper fix so we don't break it for other users. See also where this was brought up in #123, #152, #157, #182, originally implemented in #183 (and fixed at #239), re-suggested in #195 & #203, and finally more discussion an #202.

PR #312, opened 2017-10-06
Add julia multiline comment chars

RatanRSur:


alerque:
Thanks for the contribution @RatanRSur.

PR #316, opened 2017-12-26
Add support for Praat scripting language

alecshaw:
[Praat](http://www.fon.hum.uva.nl/praat/) is a program for phonetics
research which has a scripting interface.


alerque:
Thanks for the contribution @alecshaw.

PR #317, opened 2018-01-12
Add support for ANSYS Parametric Design Language (apdl)

tlsvc:
Just a new language. APDL has just one comment character, the exclamation mark ("!").


alerque:
Thanks for the contribution @tlsvc.

PR #319, opened 2018-01-18
Allow default mappings to be disabled

vasyharan:
Allow default mappings to be disabled, by default the mapping will stay enabled but allow the user to turn off the default mappings.


alerque:
Thanks for the contribution @vasyharan. I like this and it's something that's been requested before. Could I ask that you add a usage explanation of the new setting to the documentation as well before I merge this?

vasyharan:
Closing this out since `NERDCreateDefaultMappings` already exists. Read the docs. What a concept?!

alerque:
@vasyharan Ha! Even I missed that. I did think it had been requested before, I forgot somebody had implemented it.

PR #321, opened 2018-02-02
Fix Markdown Format

Sea-n:
Use `` `code` ``


alerque:
Good catch @Sea-n, thanks for the contribution.

PR #323, opened 2018-02-06
Add support for Robot Framework files

dandersson:
Add support for [Robot Framework](http://robotframework.org/) test suites.

From the [Robot Framework User Guide](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#rules-for-parsing-the-data):

> When Robot Framework parses the test data, it ignores:
>
> * ...
> * All characters following the hash character (#), when it is the first character of a cell. This means that hash marks can be used to enter comments in the test data.
> * ...

...which in general means that a line starting with `#` is regarded as a comment.

To be used with Vim filetype plugins such as https://github.com/mfukar/robotframework-vim, registering the Vim filetype `robot` .


alerque:
Thanks for the contribution @dandersson.

PR #333, opened 2018-05-25
feat(markdown): Using the most standard version of markdown commentin…

aemonge:
…g @see:

                + https://stackoverflow.com/questions/4823468/comments-in-markdown
                + https://github.com/joeyespo/grip/issues/250_


alerque:
This is not just a style difference, the actual functionality is different. HTML comments are preserved by some Markdown parsers and converted to comments in other languages. The same is not true for overloading the external reference syntax.

I can see having both styles being useful, but I don't think we should entirely replace the existing binding. Certainly adding this as an alternate delimiter would be acceptable, and maybe even adding it as the primary, but we'd need to keep the other syntax as an alternative.

Can you update this PR to not drop the current functionality?

PR #337, opened 2018-06-29
Update README.md

doc-E-brown:
Added step to build the vim help system documentation, enabling the use of :help nerdcommenter

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,8 @@ filetype plugin on
 To generate the vim help system documentation 
 
 ```sh
-vim -u NONE -c "helptags nerdcommenter/doc" -c q
+cd ~/.vim/bundle
+vim -c "helptags nerdcommenter/doc" -c q
 ```
 The vim help system can then be used to access the full documentation of all options: `:help nerdcommenter`
 
--- a/README.md
+++ b/README.md
@@ -67,7 +67,12 @@ filetype plugin on
 
 ### Documentation
 
-Please see the vim help system for full documentation of all options: `:help nerdcommenter`
+To generate the vim help system documentation 
+
+```sh
+vim -u NONE -c "helptags nerdcommenter/doc" -c q
+```
+The vim help system can then be used to access the full documentation of all options: `:help nerdcommenter`
 
 ### Settings
 


alerque:
Thanks for taking the time to contribute. I have to say I'm a little confused why you think this step is necessary at all, on all my systems having the plugin loaded also loads the help files and the help command already works as expected without this extra step.

Under what conditions would  running this command be necessary?

doc-E-brown:
Hi, I hope I can be of some help.  I am running Ubuntu 18.04 and am using Pathogen with vim, started with a clean install.  Installed Pathogen and then nerdcommenter.  I have been able to reproduce the issue in a docker container.  If you uncomment the last line of the Dockerfile, open vim and try :help nerdcommenter the error: **E149: Sorry, no help for nerdcommenter** is returned.  If you add the list line, the documentation is shown.

```
FROM ubuntu:latest

RUN apt-get update
RUN apt-get install -y vim curl git

# Install Pathogen
RUN mkdir -p ~/.vim/autoload ~/.vim/bundle && \
    curl -LSso ~/.vim/autoload/pathogen.vim https://tpo.pe/pathogen.vim

# Update vimrc
RUN echo "execute pathogen#infect()\nsyntax on\nfiletype plugin indent on" > ~/.vim/vimrc

# Install nerdcommenter
RUN cd ~/.vim/bundle && \
    git clone https://github.com/scrooloose/nerdcommenter.git

# RUN cd ~/.vim/bundle && vim -c "helptags nerdcommenter/doc" -c q
```

alerque:
I've looked into plugin managers and out of the 10 or so most used ones I see, Pathogen is the only one that doesn't generate help tags automatically. I don't want to clutter up the documentation with steps most people don't have to take. I kind of assume people will know how to use their own plugin managers. In the case of Pathogen this step is something the user is expected to do for _all_ plugins, not just this one, so it should be documented in Pathogen's README, not here.

alerque:
Please contribute this to Pathogen's readme, and then update this PR to be just a link to the generic instructions for Pathogen. I do appreciate the attempt to help, but this clutters and confuses the issue for most people that don't need this step at all, and it won't be clear who does not need to make this step.

Issue #306, opened 2017-07-27
Wrong begin/end when commenting with visual mode Ruby code


novito:
Given I have the following code:

```
def test
  puts 'test'
end

```
When I place the cursor right in the `d` of def and I do a visual selection all the way down to end, if I try to comment it, I end up with:

```
=begindef test=end
  #puts 'test'
=beginend=end
```

I am not sure why is that happening and would love to contribute if you point me to where should I be looking.

trak3r:
This seems similar to my report here https://github.com/scrooloose/nerdcommenter/issues/304

ogreface:
I've seen this too. If you've set paste, it seems to mess with it a bit more, so set nopaste before commenting. Also, I've noticed that the most consistent way to comment is c-space. Other modes have problems.

keysen:
Hello,

Same thing for me, I found this has been introduced by this commit: `1b621843051b854a3fb945cd37256e977a2a3288` 
Before this one it was properly working.

Can someone revert this commit for the vim ruby developers using this plugin please :) ?

alerque:
I've merged the temporary fix, but I'm leaving this issue open because now there is no multi-line comment function at all. That should get added back in when somebody can figure out why the Ruby syntax has gremlins.

keysen:
Thank you @alerque and yes let's wait somebody to fix the multi-line comment with block.

PR #340, opened 2018-07-24
Fix ruby comments, totally broken with =begin and =end

keysen:
It's a fix for: https://github.com/scrooloose/nerdcommenter/issues/306


ylecuyer:
I have this issue too

PR #343, opened 2018-07-27
haskell default comment style more idiomatic

goolord:
using multi-line comments is uncommon in haskell, and it makes more sense to use multiple single line comments rather than multiple mult-line comments. this changes the default behavior from
```haskell
{- foo :: Int -> Int -}
{- foo a = a + 1 -}

{- bar :: Text -> Text -}
{- bar x = x <> "text" -}
```

to

```haskell
-- foo :: Int -> Int
-- foo a = a + 1

-- bar :: Text -> Text
-- bar x = x <> "text"
```

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -206,7 +206,7 @@ let s:delimiterMap = {
     \ 'h': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'haml': { 'left': '-#', 'leftAlt': '/' },
     \ 'handlebars': { 'left': '{{!-- ', 'right': ' --}}' },
-    \ 'haskell': { 'left': '--', 'nested': 1, 'leftAlt': '{-', 'rightAlt': '-}', 'nestedAlt': 1 },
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for looking into this. It does look like the single vs. multiline configs got reversed for Haskell. That being said I'm not quite sure this is right yet. Shouldn't the default single line format _not_ have nesting enabled?

goolord:
That makes sense to me, fixed.

goolord:
Sure:

-  According to the most popular haskell style guide, [tibbe/haskell-style-guide](https://github.com/tibbe/haskell-style-guide/blob/master/haskell-style.md#comments), comments spanning multiple lines use single line comments.
- Multi-line comments are not even mentioned on the (admittedly outdated) [haskell wiki](https://wiki.haskell.org/Commenting). 
- An example of this from pandoc (the most popular package on hackage)
```
--
-- Date/time
--

-- | Parse a date and convert (if possible) to "YYYY-MM-DD" format. We
-- limit years to the range 1601-9999 (ISO 8601 accepts greater than
-- or equal to 1583, but MS Word only accepts dates starting 1601).
```
- It doesn't make much sense to me to comment out individual lines with individual multi-line comments

alerque:
Thanks for taking the time to contribute @goolord.

PR #355, opened 2018-12-23
make g:NERDToggleCheckAllLines not check blank lines

zhanglintc:
Hi, I'm not sure this is a bug or not, but I think my pull request would make a better performance.

Currently, when I set "g:NERDToggleCheckAllLines = 1", if I use `<leader>c<space>`, it would check all my selected lines(including blank lines!), if one of theses lines is not commented, it will comment all lines.

For example:

```Python
# this is the first line

# this is the third line
```

Note that the second line is blank. If I select line 1 to 3, and press `<leader>c<space>`, then the code will become:

```Python
# # this is the first line

# # this is the third line
```

While what I want to do is uncomment line 1 and line 3. I don't think somebody woud like to comment line 1 and line 3 again.

So I made this pull request to let "g:NERDToggleCheckAllLines" not check the blank lines. It should only check the valid lines(I think blank line is not valid lines).

If you think what I changed is better, please accept my pull request.

If you have some other reasons that "g:NERDToggleCheckAllLines" should check all lines include blank lines please tell me why. 

Thank you.

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -1270,8 +1270,8 @@ function! NERDComment(mode, type) range
           let l:commentAllLines = 0
           for i in range(firstLine, lastLine)
             let theLine = getline(i)
-            " if have one line no comment, then comment all lines
-            if !s:IsInSexyComment(firstLine) && !s:IsCommentedFromStartOfLine(s:Left(), theLine) && !s:IsCommentedFromStartOfLine(s:Left({'alt': 1}), theLine)
+            " if have one line no comment(not include blank/whitespace-only lines), then comment all lines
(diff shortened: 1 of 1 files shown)


alerque:
This sounds like a good change to me. I'd be interested in hearing if other active users have a use case where this would not be desirable.

In the mean time I would change your check to also check for lines that are only whitespace as some editors & languages maintain whitespace indent levels even on otherwise "blank" lines. It seems like these should be processed the same way, no?

zhanglintc:
Sounds reasonable, it would be better to ignore lines only contain whitespaces, too.
I think it could be something like:

`if theLine =~ "^[ \t]+$" && XXXXXXX && XXXXXXX`

Need I create another pull request?

alerque:
@zhanglintc No, no need for another pull request. You can make another commit in the same branch and add it to this PR by pushing to the same place. You _could_ also squash it with the last commit and force-push to combine them, but I can also do that when I merge the PR so if you don't know how to do it don't worry about it. Thanks for contributing.

alerque:
Thanks for the contribution @zhanglintc, and especially for adapting the requested changes. Sorry this PR fell off my radar for a while, it's been merged now.

PR #359, opened 2019-01-31
updated Django comments

### LINKED ISSUE #358: Update htmldjango Comment
The way to comment in django template is  through the `{% comment %} code {% endcomment %}` not the usual html way which was implemented

MeNsaaH:
Fixed #358 . This PR just adjusts the django comments.


plugin/NERD_commenter.vim
```diff
@@ -144,7 +144,7 @@ let s:delimiterMap = {
     \ 'desktop': { 'left': '#' },
     \ 'dhcpd': { 'left': '#' },
     \ 'diff': { 'left': '#' },
-    \ 'django': { 'left': '<!--', 'right': '-->', 'leftAlt': '{#', 'rightAlt': '#}' },
+    \ 'django': { 'left': '{% comment %}', 'right': '{% endcomment %}', 'leftAlt': '{% comment %}', 'rightAlt': '{% endcomment %}' },
```
alerque: If there is not an alternative comment syntax available, please just don't set the alternate delimiters.

```suggestion
    \ 'django': { 'left': '{% comment %}', 'right': '{% endcomment %}' },
```

plugin/NERD_commenter.vim
```diff
@@ -172,7 +172,7 @@ let s:delimiterMap = {
     \ 'exports': { 'left': '#' },
     \ 'factor': { 'left': '! ', 'leftAlt': '!# ' },
     \ 'fancy': { 'left': '#' },
-    \ 'faust': { 'left': '//' }, 
+    \ 'faust': { 'left': '//' },
```
alerque: Please leave whitespace cleanup for a separate commit, otherwise `git blame` starts giving useless results. If in the future something comes up about faust and I want to know what it is, what put it there, why we used the delimiters we  did, etc, I will `git blame` this file to find who contributed it and what issuer or PR it was discussed in. Mucking around through unrelated commits is no fun. Thanks.

plugin/NERD_commenter.vim
```diff
@@ -216,7 +216,7 @@ let s:delimiterMap = {
     \ 'hog': { 'left': '#' },
     \ 'hostsaccess': { 'left': '#' },
     \ 'htmlcheetah': { 'left': '##' },
-    \ 'htmldjango': { 'left': '<!--', 'right': '-->', 'leftAlt': '{#', 'rightAlt': '#}' },
+    \ 'htmldjango': { 'left': '{% comment %}', 'right': '{% endcomment %}', 'leftAlt': '{% comment %}', 'rightAlt': '{% endcomment %}' },
```
alerque: Is this an embedded language used inside of HTML? If so the alternate comment syntax should be for the language it co-exists with, perhaps something like this:

```suggestion
    \ 'htmldjango': { 'left': '{% comment %}', 'right': '{% endcomment %}', 'leftAlt': '<!--', 'rightAlt': '-->' },
```

plugin/NERD_commenter.vim
MeNsaaH: Sure. This must have being done from my vim, it automatically trims whitespace. But will surely revert it

MeNsaaH:
The changes have being implemented.

plugin/NERD_commenter.vim
MeNsaaH: An alternate comment used amongst django community members is `{# code #}` which I've updated

alerque:
Thanks @MeNsaaH, that makes a lot more sense with the alternate syntax for inline usage. Thanks for the contribution.

MeNsaaH:
I'm glad I did

PR #360, opened 2019-02-11
Add spaces to haskell comments

5outh:
Using `nerdcommenter` to comment out Haskell code is problematic because `--` pre-pended to a bunch of symbols is syntactically valid code.

For example, the line:

```haskell
  -> a
```

gets commented as

```haskell
  ---> a
```

and `--->` can be defined as an operator in haskell, so the line isn't commented out. This commit adds a space before line comments and spaces between alt comments for consistency.


alerque:
This should be done by setting a preference in your RC file:

```vim
let g:NERDSpaceDelims = 1
```

We don't have a proper mechanism for forcing this on some file types, but we did run into some problems having spaces as part of the delimiters (especially unexpected results for people who have spaces turned on anyway).

I appreciate the contribution, but I can't merge this as it is without messing up other people. I'd be happy to accept something that forced that preference to be on for file types that need it!

5outh:
@alerque The preference is enough for me. Thanks for pointing me in the right direction!

PR #354, opened 2018-12-21
Add support for Faust

glegodais:
Faust is a functional programming language for sound synthesis and audio processing: http://faust.grame.fr/

The file extension is **_.dsp_** and the vim filetype is **_faust_**

PR #365, opened 2019-04-02
Add support for Faust

alerque:


alerque:
Oops, duplicate of #354. The Github network map is not always as informative as it could be!

PR #368, opened 2019-04-02
Make plugin autoload and refactor accordingly

alerque:
I was perusing the network map a bit and ran across these changes in @Houl's fork. Not all of them are suitable for merging here, but at least some of this work should be considered for cleanup and upstreaming. This PR is to get that process started.


README.md
```diff
@@ -1,159 +1,25 @@
...
-    Adds comment delimiters to the end of line and goes into insert mode between them.
-
-  * **|NERDComInsertComment|**
-
-    Adds comment delimiters at the current cursor position and inserts between. Disabled by default.
-
-  * `<leader>ca` **|NERDComAltDelim|**
-
-    Switches to the alternative set of delimiters.
+For initial instructions, check out the original script at
```
alerque: In order to consider this PR for upstreaming can we revert these README changes and anything else that doesn't belong being merged upstream? If you want to switch to doing this on a branch instead of your fork's master that's fine.

README.md
```diff
@@ -1,159 +1,25 @@
...
+[https://github.com/scrooloose/nerdcommenter](https://github.com/scrooloose/nerdcommenter)
 
-  * `[count]<leader>cl`  
-    `[count]<leader>cb` **|NERDComAlignedComment|**
+## Configuration
 
-    Same as **|NERDComComment|** except that the delimiters are aligned down the left side (`<leader>cl`) or both sides (`<leader>cb`).
+Suggested mappings to be put in the vimrc:
```
alerque: Can we make any changes to default bindings part of a different PR (if it's even still applicable, this seems like something that should be part of a user configuration and shouldn't have to be part of a fork at all).

README.md
Houl: Ok, I'll try that.  Any branch name suggestions?

README.md
Houl: Do you still want a plugin file with default mappings or not?  If yes, my README's mapping suggestions could go there.  If not, how about adding the mapping suggestions to the original README?  In any case, the first commit in the new branch will contain the original README unchanged.

README.md
alerque: I'd just call it `autoload`. Then I'd rebase it on upstream/master and drop the any unrelated commits and open a new PR. I'll close this one when you do since it's against your `master`.

Houl: Branch created.

alerque: Great! Can you [open it as a PR too](https://github.com/scrooloose/nerdcommenter/compare/master...Houl:autoload?expand=1? That way it will show up as being your contribution.

alerque:
Replacing with #376.

PR #292, opened 2017-03-25
Add support for piglit's shader_test files

Lyude:
This adds support for shader_test files in piglit, the open source
test suite for OpenGL implementations such as mesa. More information on
piglit can be found at:

https://piglit.freedesktop.org/

PR #371, opened 2019-04-02
Add support for piglit's shader_test files

alerque:
This adds support for shader_test files in piglit, the open source
test suite for OpenGL implementations such as mesa. More information on
piglit can be found at:

https://piglit.freedesktop.org/


alerque:
Oops, duplicate of #292. The Github network map is not always as informative as it could be!

PR #376, opened 2019-05-01
Refactor using VIM's autoload plugin system

Houl:


alerque:
Great, thanks for opening the PR. I can work on cleaning up this branch here by rebasing it etc.

alerque:
A long time coming, but I've finally rebased this branch to remove the things we don't want (such as stripping the default mappings), squash some of the little fixup commits, remove duplicate merges, and finally merge the latest master branch.

Now all it needs is testing! Anybody?

antonk52:
Is this PR still valid? Happy to test or help get it merged

alerque:
Yes it should still be valid. I see it needs another merge, probably to catch a few new file types. What it really needs is testing on a few different setups such as VIM and NeoVIM using both a manual install method and one or two plugin managers in each case to make sure everything is where it is expected to be. @anatonk52 if you can help with any of that I'll definitely facilitate. Sorry I haven't gotten around to it on my lonesome.

antonk52:
Thanks @alerque ! In the next few day I will try it out using [neo]vim and report problems if any.

alerque (approved):

alerque:
I've given this a shake in my personal setup (a highly eccentric NeoVIM setup that is *not* representative of normal was people might use plugins) and it works fine so I think it's probably okay. If you test it in more generic (Neo)VIM arrangements without crazy custom configs and it works as expected out of the box then I'd say this is good for merge, feel free to do the honors.

antonk52:
Hi @alerque 

I've tried to test out the [`autoload`](https://github.com/Houl/nerdcommenter-vim/tree/autoload) branch. What I found is that the main file has moved to autoload folder. Which means it is not sourced during the [neo]vim startup(good). This also means that no initial keymaps are set so the users would have to source this file themselves(suboptimal).

The common practice I found in popular plugins and writing my own is to have both an `autoload/*.vim` and `plugin/*.vim` file where the initial keymaps are set which in turn call autoloaded functions. This way only the absolutely necessary work is done during the startup and the rest is being loaded once called for it via autoloaded functions. By doing so it is also possible to preserve backward compatibility with all vim versions supporting autoloaded functionality therefore the users after the update would not need to tinker with their setup which is a plus.

Besides the need of explicitly loading the plugin by calling any autoloaded function or referring for a variable the commenting works as expected. I was able to test it using both Neovim 0.5 (with custom setup and plain) and plain vim 8.2. Hope this helps.

alerque:
Yes that help.

Any chance you'd be able to figure out the binding bits that need to be moved back to a regular plugin location? You should have push access to this branch.

antonk52:
Oh, I didn't realise I have push access. I should be able to separate the lazy stuff from the essential keybindings. I will give it a go over the weekend 👍

antonk52:
The summary of my changes:
- setup global variables in `plugin/nerdcommenter.vim` so they are available within user's scripts without having to call autoloader functions.
- map commands in `plugin/nerdcommenter.vim`.
- `SwitchToAlternativeDelimiters` is now an autoloaded function so `<SID>` no longer necessary.
- removed autocommands on bufread/bufenter for `nerdcommenter#SetUpForNewFiletype` which is now called on each call to `nerdcommenter#SwitchToAlternativeDelimiters` and `nerdcommenter#Comment` functions. It also tried to avoid unnecessary work if it was already setup for current filetype.

I have tested the startup time to see how much startup time we were able to shave off and on my current machine the current version of the plugin takes 3.5ms and the new version with autoloaded functions takes 1.6ms

The only drawback I found is that the delimiter map has moved to the autoloaded directory which means `g:NERDDelimiterMap` is not being set until anything from the autoloaded directory is called. I have added `nerdcommenter#delimiterMap` for users to have an option to have access to the delimiter map without having to call any functions.

I have tested this in neovim 0.5 and vim 8.2 and things seem to work as expected. Feel free to give it a try.

alerque:
Thanks  a lot! This is a long overdue overhaul. I gave this a shot in my own setup and it worked fine. I tweaked the function definitions to make the linter happy and will merge now.

PR #377, opened 2019-06-02
Update documentation for 11 year old NERDCommenter[A-Z] style

jamespharvey20:
11 years ago, b54d3bb renamed `NERDCom[A-Z]` mappings to `<plug>NERDCommenter[A-Z]` mappings.  (Unknown if intentional.)  But, the docs were never updated.  When README.md was later created, it kept the old `NERDCom[A-Z]` names.

Without these changes, it can be quite confusing to someone trying to setup their own mappings, and having nothing work!  Found a few previous issues that look like they were caused by this confusion.

I don't know if splitting the `NERDCommenterAlign{Left,Both}` tags as in the second commit is customary, but it works and still avoids having two separate help sections.  I don't feel replacing `NERDComAlignedComment` with `NERDCommenterAlignedComment` is the proper fix here, because someone searching for a `NERDCommenterAlign{Left,Both}` tag isn't going to find it.

Also update other help tags from `NERDCom[A-Z]` pattern to `NERDCommenter[A-Z]` for consistency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,57 +103,57 @@ let g:NERDToggleCheckAllLines = 1
 
 The following key mappings are provided by default (there is also a menu provided that contains menu items corresponding to all the below mappings):
 
-Most of the following mappings are for normal/visual mode only. The **|NERDComInsertComment|** mapping is for insert mode only.
+Most of the following mappings are for normal/visual mode only. The **|NERDCommenterInsert|** mapping is for insert mode only.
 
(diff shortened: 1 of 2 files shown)


alerque:
Thanks for your work on cleaning this up. Now that you bring it up I remember being confused about this myself when I first started using the plugin.

I'll need to review this a bit and make sure there are no unintended side affects, but my first impression was that this will be fine.

PR #390, opened 2019-08-06
Add proto filetype to support protobuf.

ufengzh:


alerque:
Thanks for the contribution @ufengzh.

Issue #280, opened 2017-01-09
React - jsx


xavierartot:
Hi,
I'm a huge fan, but with React and jsx it doesn't work.
Thanks

alerque:
Are they listed in the file type matrix? Do you mean the file types aren't supported yet or they aren't working _properly_?

chrisnojima:
Hi,

Maybe I'm misunderstanding how the vim filetype declaration works but I have my .js files set as javascript.jsx so I can get js and jsx formatting.  These files have a mix of plain js and blocks of jsx (typical react files) yet with this change all comments are of the form `{/* to comment out */}`. But this doesn't work on plain js. For example i see this

```
// This is correct
const render = () => (
  {/* <div /> *}
)

// This is incorrect
{/* const a = 'hello' */}

```

alerque:
@chrisnojima Vim doesn't provide (at least not readily) context sensitive information about what type of language block you are it. As far as NerdCommenter is concerned this is pretty much a one shot deal. Either the filetype is plain JS (in which case you get  plain JS comments) or it is JSX, in which case you  get React style comments. It doesn't try to parse the surrounding lines to toggle back and forth between language modes.

There are several languages for which this would be desirable, but I'm not aware of a good way to implement this yet. I would suggest opening this as a separate issue (if we don't  have one already) as a potential improvement. I'd be happy to see a contribution to this effect.

In the mean time I don't thing you have an easy option to handle this other than  changing the file type or overriding the JSX comment type to generate plain JS comments.

amadeus:
Ok, IMO the fix that got merged in for this should be reverted.

JSX files are just Javascript really, with some specialized syntax every once in a while. This current change breaks commenting completely for JSX files. My vote would be to revert this change and just force people to comment manually within JSX scope.

Take this super simplified example:


```jsx
import React, {PureComponent} from 'react';

class MyComponent extends PureComponent {
    handleClick() {
        console.log('You\'ve clicked the element!');
    }

    render() {
        return (
            // The only actual JSX witin the file
            <div onClick={() => this.handleClick()}>Example JSX syntax</div>
        );
    }
}

export default MyComponent;
```

It's now impossible to use NERDCommenter anywhere within the file, EXCEPT within that single one line in the return statement.

argshook:
totally agree with @amadeus, this change took me by surprise. I very rarely comment actual jsx code.

[...]

PR #393, opened 2019-08-17
Change alternate style for .jsx / .tsx filetypes

impulse:
Changed based on following solution to have comments in jsx / tsx: https://github.com/scrooloose/nerdcommenter/issues/280#issuecomment-395566184

I would prefer this to be the default option. Just thought I'd leave this here :)


plugin/NERD_commenter.vim
```diff
@@ -237,6 +237,7 @@ let s:delimiterMap = {
     \ 'java': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'javacc': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'javascript': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
+    \ 'javascript.jsx': { 'left': '//', 'right': '', 'leftAlt': '{/*', 'rightAlt': '*/}' },
```
alerque: Is this a standard vim `filetype` code?

impulse: @alerque Thanks for pointing that out (didn't check on that). Apparently this is non standard, however I've seen many plugins requiring setting these in order to function (other editors like vscode seem to use it too). I can totally understand if that is an issue though.

alerque: Links? A few other major plugins that require this might be precedent enough. If we start adding non-vim standardized types to this list though there are a couple things I'd want to see changed. First I think I'd keep them in a separate array in the code or somehow clearly partitioned off so that it was clear what was vanilla filetypes and what is hacks. Secondly I think there should be some documentation explaining that some of the "supported" types need to be manually set as they are not things vim will set. Somewhere in the plugin setup should tip people off to not all file types being created equal.

Lastly if there are any discussions about _adding_ these file types to vim/nvim that would be immanently relevant too.

impulse: I assume all plugins trying to fully support Microsofts language server protocol are affected by this. As you can see in the [LSP specification](https://microsoft.github.io/language-server-protocol/specification) there are filetypes named `javascriptreact` and `typescriptreact`. This is also the case for [VSCode](https://code.visualstudio.com/docs/languages/identifiers) since it is built upon that. 

Taking a look at my autocomplete plugin of choice (coc.nvim), they also seem to require to set either of those filetypes f.ex [(javascriptreact or javascript.jsx)](https://github.com/neoclide/coc.nvim/blob/b9967440b806a20067342b2ecc09682e6e1c685c/src/model/document.ts#L86) in order to have support for jsx/tsx specific snippets. This is also confirmed by a maintainer [here](https://github.com/neoclide/coc.nvim/issues/459#issuecomment-465658887).

Couldn't find an issue in the vim/nvim repos discussing this, so I might just create an issue over there. This would be the best way to avoid anything hacky / unofficial.

alerque:
I'm quite willing to accept a stand in ad-hoc filetype if we need to, but given there is a response from the primary vim maintainer on this issue I think I'll hold off a few mare days at least in case we actually get a bonified standard filetype added upstream.

impulse:
@alerque seems we got `javascriptreact` and `typescriptreact` accepted https://github.com/janlazo/neovim/commit/f667a0e02ac5781a7b7e640b5dc9a7edcf2f4a0b 🎉. 

It might still be a good idea to also keep `javascript.jsx` and `typescript.tsx` until other filetype-specific plugins adapt to the change. I'll update the PR if you want me to.

alerque:
That's awesome @impulse thanks for seeing that process through.

If we had the hack for the dot*sx variations released I would say wait to pull it out until at least the next major release cycle on Vim/Neovim, but since we dot't have it it place yet and many React folks will already have their configs tweaked anyway, I think I'd like to just add the final filetype values to this plugin. I don't see any need to perpetuate what is going to be a dead filetype.

Can you rework this using `javascriptreact` and `typescriptreact`? It might not hurt to add a note in the docs / readme about _why_ we chose that for React support.

As for any other plugins that use the dot*sx filetypes lets encourage them to at least add a mapping to the now official filetype and start porting people that have adopted the old hack over.

impulse:
Gotcha, updated for the official vim filetypes then. 

I am not a native speaker but a possible explanation for the docs could be:

**React**: Commenting within your jsx or tsx markup requires comments to be wrapped in curly braces.
Thus vim's official `javascriptreact` and `typescriptreact` filetypes will use a different alternative set of delimiters.

plugin/NERD_commenter.vim
```diff
@@ -237,6 +237,7 @@ let s:delimiterMap = {
     \ 'java': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'javacc': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'javascript': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
+    \ 'javascriptreact': { 'left': '//', 'right': '', 'leftAlt': '{/*', 'rightAlt': '*/}' },
```
alerque:
```suggestion
    \ 'javascriptreact': { 'left': '//', 'leftAlt': '{/*', 'rightAlt': '*/}' },
```

plugin/NERD_commenter.vim
```diff
@@ -451,6 +452,7 @@ let s:delimiterMap = {
     \ 'twig': { 'left': '{#', 'right': '#}' },
     \ 'txt2tags': { 'left': '%' },
     \ 'typescript': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
+    \ 'typescriptreact': { 'left': '//', 'right': '', 'leftAlt': '{/*', 'rightAlt': '*/}' },
```
alerque:
```suggestion
    \ 'typescriptreact': { 'left': '//', 'leftAlt': '{/*', 'rightAlt': '*/}' },
```

PR #397, opened 2019-10-05
Add syntax for troff family languages

roryrjb:
This change explicitly adds mappings for troff family languages: troff, groff and mandoc. This _could_ be mostly covered by the current mapping for __man__, but in the case of troff and mandoc `.\"` is preferred as it avoids additional whitespace (https://mandoc.bsd.lv/man/roff.7.html#Comments) and similarly in groff `\#` is used (https://www.gnu.org/software/groff/manual/html_node/Comments.html) for the same reason.

Also, as in my case, it helps for those who might have `*.1`, etc files mapped to either one of troff, groff or mandoc.

And finally although man is in troff, both troff and groff are used for other things too.


alerque:
Looks great to me @roryrjb, and the explanation makes sens. Thanks for the contribution!

roryrjb:
@alerque many thanks!

Issue #400, opened 2019-10-20
Migrate repository to shared org namespace


alerque:
Since sometime in 2016 (see #236) I've been the ~~primary~~ sole maintainer of this plugin, the only developer actively managing the repository, helping PR's to get merged, attending to bugs, and tinkering with new features. It hasn't been a huge issue for me to have it in another namespace and I certainly haven't wanted it in my personal namespace because, having not developed in in the first place, I don't feel like I am the owner.

However as time goes on more issues come up that make it hard to maintain a plugin without any ownership of the repository. Github is constantly adding features, and I could do a lot to help the PR workflow if I had more access. The repository owner remains completely uninvolved and also unreachable, making it impractical to ask him for every change.

Additionally I've run across more good plugins that have stagnated because their initial developers have moved on to other things. The community continues to use them and even finds and fixes bugs, but the PRs just build up because the repository owners are elsewhere. I've started taking care of a few of these, including `vim-ledger`, `vim-pandoc`, and others. The latest one I ran across and picked up maintenance for is [tagvim](https://github.com/majutsushi/tagbar/issues/549).

I think it is time to move this plugin to a namespace where I have some access to tweak the settings, setup CI jobs, etc. This will avoid the appearance of trying to take over 'ownership' (credit still goes to @scrooloose for creating this plugin) but give some leeway to maintain it properly. I think this will also encourage community contributions in the end as it will be more apparent that it is a group effort, not one person's baby.

To that end I have created the an org namespace @preservim where this and possibly later other plugins I'm helping to maintain could be hosted. **I would like to request this plugin be migrated to that namespace.** @scrooloose I have invited you to the org. Please let me know how this idea sits with you.

Note that migration will not break any of the existing links, credits, git clones, or anything because GitHub keeps redirects from the old namescpace.

wsdjeg:
see my previous comment, https://github.com/majutsushi/tagbar/issues/549#issuecomment-544237548

alerque:
@wsdjeg Your comment on the other project only suggested that the original namespace would be "easier for searching", but I noted that GitHub automatically adds redirects from the old namespaces so that search results and even existing git clones do not lose their connection to the project, everything follows the migration pretty smoothly. I can't think of any downside related to searching, and you didn't respond there. Did you still have a concern related to this?

[...]

PR #401, opened 2019-10-25
Add VIM built-in package management to the read me

pesarkhobeee:
Dear @scrooloose 
I just updated the README file in order to cover the new feature of VIM version 8 which is built-in package management. hope it brings more ease for new users of your awsome plugin :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,20 @@ Comment functions so powerful—no comment necessary.
 
 ## Installation
 
+Below are just some of the methods for installing nerdcommenter. Do not follow all of these instructions; just pick your favorite one. Other plugin managers exist, and nerdcommenter should install just fine with any of them.
+
+#### Vim 8+ packages
+
+If you are using VIM version 8 or higher you can use its built-in package 
+solution, remember, in this case you should keep the plugin update manuelly, see `:help packages` for more information. Just run these commands in your terminal:
+
+```bash
+git clone https://github.com/scrooloose/nerdcommenter.git ~/.vim/pack/vendor/start/nerdcommenter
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,19 @@ Comment functions so powerful—no comment necessary.
 
 ## Installation
 
+Below are just some of the methods for installing nerdcommenter. Do not follow all of these instructions; just pick your favorite one. Other plugin managers exist, and nerdcommenter should install just fine with any of them.
+
+#### Vim 8+ packages
+
+If you are using VIM version 8 or higher you can use its built-in package management; see `:help packages` for more information. Just run these commands in your terminal:
```
alerque: As far as I understand it this is technically inaccurate. VIM 8 doesn't mange packages, it just gives you a place where you can manage them yourself. See for example [minpac](https://github.com/k-takata/minpac) which specifically wraps this feature with actual package management.

I think I would reword this section including the title so the distinction between manually cloning (and having to manually update) vs. package managers that fetch and update themselves. The issue here is less 3rd party vs. 1st party as it is the kind of function is manual vs. automatic.

alerque (changes requested):
Thanks for the contribution @pesarkhobeee! The original author of this hasn't been active for many years now, I've been managing it for 3 years or so and it was stagnant for a long while before that (see #400 for current status), but yes the author did setup a great plugin.

I agree some instructions for installing manually into plain VIM 8's package path are in order, but I think the actual wording here kind of muddies the water of what a plugin manager actually does. Can you review my notes on that and update accordingly?

README.md
pesarkhobeee: Dear @alerque 
Thanks for your detailed description and sorry for my late replay, I was on a trip
I got your point and I totally agree, I will change it

Issue #402, opened 2019-10-26
backslashes aren't handled consistently


roryrjb:
I added a pull request recently (https://github.com/scrooloose/nerdcommenter/pull/397) that was merged, this added support for troff-family (groff, troff, mandoc) comment delimiters, related to, but distinct from the existing `man`/manpage delimiters. This worked for adding the delimiter, but I've only just noticed that when trying to remove the delimiter it doesn't work and just adds another delimiter, i.e.:

```troff
.\" with this comment line I press <leader>C <space>
.\" .\" I get this
```

...I'm not _that_ familiar with either this code base or Vim scripting but I believe it's just how backslashes are handled. From looking at all the delimiter definitions, the recent pull request above was the first time backslashes were present so I believe this has never come up.

roryrjb:
I have opened a pull request to try to address this: https://github.com/scrooloose/nerdcommenter/pull/403.

PR #403, opened 2019-10-26
Improve handling of delimiters with backslashes

roryrjb:
This pull request has been opened to address issue https://github.com/scrooloose/nerdcommenter/issues/402 I raised separately.

As mentioned in the issue description, I recently added delimiter definitions for troff family languages, which included, for the first time backslashes:

```viml
\# 	" groff
.\" 	" troff / man doc
```

...but these aren't handled correctly in some cases. Currently the plugin can add them but they won't be detected when trying to remove them, so they get added again.

This pull request is trying to solve this, although it might not be in the best way and others may have a better solution to this, but of course if I can contribute then I'll try to.

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -2070,6 +2070,14 @@ function s:FindDelimiterIndex(delimiter, line)
     "get the index of the first occurrence of the delimiter
     let delIndx = stridx(a:line, l:delimiter)
 
+    "if there is no match and the delimiter contains two backslashes
+    "then try indexing again but replace those backslashes with only a
+    "single backslash on the assumption that the two backslashes really
+    "are an escape for a single backslash
+    if delIndx == -1 && stridx(l:delimiter, '\\') > 0
+        let delIndx = stridx(a:line, substitute(l:delimiter, '\\\\', '\', ''))
+    endif
+
     "keep looping thru the line till we either find a real comment delimiter
     "or run off the EOL
     while delIndx != -1


roryrjb:
Just wanted to add that the definition for troff/mandoc probably should be `.\"` rather than the `.\\"` currently and this pull request relies on this.

Also this pr doesn't actually fix the issue with groff `\#`. Again someone else might have a better solution for this, but I will try to come up with some additional fixes in the meantime. If I do get something together that is better I will tidy up and squash this pr.

alerque:
Thanks for the report and the attempt to fix ti @roryrjb. I'm not entirely sure myself how this should be fixed. Also I won't have time to dig into it for at least a few days.

As I understand it this PR currently fixes troff but not groff right?

Also I would suggest going ahead and adding a commit to this PR that changes the default delimiter for troff/mandoc if you think you can make the case for the default change being for the best.

roryrjb:
@alerque that's correct, this only fixes troff but I'm going to see if I can fix groff too. I'll also see if changing the delimiters for troff and mandoc will improve things, because I think that will make them closer to groff and all can be fixed in the same way.

alerque:
What's the status on this @roryrjb? Let me know when I should give this another review. Sorry I wasn't very available back when it came in, if you need something let me know.

roryrjb:
@alerque hey, I haven't had much time but still want to crack this, especially as I'm working with troff this weekend. I'm hoping to have a fix very soon, will let you know. Hope you had a good Christmas and all the best for 2020!

PR #406, opened 2019-11-20
Adds New Delimiter for Hashicorp Sentinel Files

ncronquist:
- Sentinel is a new language developed by Hashicorp
- Sentinel supports three types of comments, `#` and `//` for single
  line comments and `/* ... */` for multi-line comments as described
  in their documentation - https://docs.hashicorp.com/sentinel/language/spec/#comments
  - Since sentinel and terraform, .tf extension, are related, the `#`
    symbol was chosen to keep in line with the tf single line comment


alerque:
Thanks for the contribution and explanation @ncronquist.

PR #408, opened 2019-12-21
Add semantic patch (.cocci) to delimiterMap

jajajasalu2:
Coccinelle is a tool used for transforming C source code. It
uses the Semantic Patch Language (.cocci files) for doing this.
The conventional notation for commenting in semantic patches
is "//".

Used in conjunction with this plugin:

https://github.com/ahf/cocci-syntax

The cocci filetype is set, following which NERD_commenter can
be used in .cocci files.

Signed-off-by: Jaskaran Singh <jaskaransingh7654321@gmail.com>


alerque:
Thanks for taking the time to contribute @jajajasalu2, and especially for the detailed explanation of what this is, what plugin(s) set the `filetype`, and such. That makes my job as a reviewer easy, and this looks great.

jajajasalu2:
No problem, thanks for merging!

PR #411, opened 2019-12-30
Create LICENSE file for CC0

### LINKED ISSUE #217: Licence change request
Hi,
I respect, that the author has the right to select his favourite license but this license:
https://github.com/scrooloose/nerdcommenter/blob/master/plugin/NERD_commenter.vim#L7
makes a distribution complicate. Public domain is complicate in general [1], but especially this version:
It requires a rename of the project. 
It would be very helpful for all Linux distributions and users, if you would select one of the suggested open source licenses[2]
Found out about the problem due to a bug report on gentoo.
Best,
JS

[1] http://opensource.org/faq#public-domain
[2] http://opensource.org/licenses/alphabetical
[3] https://bugs.gentoo.org/show_bug.cgi?id=566430

alerque:
Closes #217. I'll leave this PR open for a while waiting for any other feedback on that issue or here, but as of my current review I think this would be my vote going forward.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,121 @@
+Creative Commons Legal Code
+
+CC0 1.0 Universal
+
+    CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE
+    LEGAL SERVICES. DISTRIBUTION OF THIS DOCUMENT DOES NOT CREATE AN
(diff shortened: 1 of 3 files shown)


wsdjeg:
please keep it same as before.

alerque:
@wsdjeg Why? What is the benefit of the current license to you or the context you use it in that would not be provided by my suggested replacement? If you have a case to make for this please respond to my issue discussing the pros and cons.

PR #416, opened 2020-01-21
Feature: API 'NERDCommentIsCharCommented(line, col)'

li-zyang:
Added an API so that we can use NerdCommenter plugin to check if a character is inside a comment or not
Useful for developing plugins depending on NerdCommenter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *~
 *.swp
 tags
+DEBUG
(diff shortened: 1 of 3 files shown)


alerque:
Thanks for this great contribution @li-zyang, and a special thanks for documenting it well along the way.

li-zyang:
Well...I'd just found that the ```|...|``` around ```NERDCommenterInterfaces``` in the catalog of the documention is missing. My color scheme hides these symbols in the help files so I didn't noticed it. You may fix it in the document.
```
    5. Customising key mappings...............|NERDCommenterMappings|
    6. Interfaces............................. NERDCommenterInterfaces
    7. Issues with the script.................|NERDCommenterIssues|
        7.1 Delimiter detection heuristics....|NERDCommenterHeuristics|
        7.2 Nesting issues....................|NERDCommenterNesting|
```

alerque:
Done, thanks.

PR #420, opened 2020-03-01
fix haskell left comment

chrissound:
`--` should be `-- `

Also fixed here https://github.com/preservim/nerdcommenter/pull/96 (but it's old + conflicts)


alerque:
@chrissound Adding spaces to comment delimiters where they are not _strictly required_ by the language's syntax creates quite a few problems. Instead please use the following setting if using the space is your preference:

```vim
let g:NERDSpaceDelims = 1
```

PR #442, opened 2020-10-30
Add support for the `nix` language.

syberant:
This PR adds support for the nix language, it just uses a single `#` on the left to comment out a line so I've added this to the `delimiterMap`.
The nix language is used by [nixpkgs](https://github.com/NixOS/nixpkgs), [NixOS](https://nixos.org/) and some other Nix-related projects.

Thank you for the plugin! Just trying to get this in so future nix programmers don't have to set `NERDCustomDelimiters`.


alerque:
Thanks for taking the time to send in a contribution!

PR #444, opened 2020-11-18
Add NERDUseDefaultMappings variable to be able to set own mappings from scratch

uWynell:
When I first installed this extension, I wanted to set up my own mappings from scratch, rather than using existing ones.
And I just put up with it...
But now I've decided to try my luck and do this PR.
This commit doesn't change anything for people at all.
It only adds the ability not to use default mappings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,9 @@ Please see the vim help system for full documentation of all options: `:help ner
 Several settings can be added to your vimrc to change the default behavior. Some examples:
 
 ```vim
+" Use default mappings (set to 0 to set own from scratch)
+let g:NERDUseDefaultMappings = 1
+
 " Add spaces after comment delimiters by default
 let g:NERDSpaceDelims = 1
 
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -34,6 +34,7 @@ let s:spaceStr = ' '
 let s:lenSpaceStr = strlen(s:spaceStr)
 
 " Section: variable initialization
+call s:InitVariable('g:NERDUseDefaultMappings', 1)
 call s:InitVariable('g:NERDAllowAnyVisualDelims', 1)
 call s:InitVariable('g:NERDBlockComIgnoreEmpty', 0)
 call s:InitVariable('g:NERDCommentWholeLinesInVMode', 0)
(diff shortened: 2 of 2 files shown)


uWynell:
I'm actually stupid, sorry
```vim
g:NERDCreateDefaultMappings
```
already exists
I just did not see it in a README and missed in the source code

alerque:
Umm, did you check the docs first? I think this already exists:

```vim
let g:NERDCreateDefaultMappings = 0
```

Additionally all of the default mappings only actually get run if the same function isn't already bound to something else, so you can create custom bindings to some functions and skip the default ones while still getting the other defaults.

alerque:
...do appreciate the contribution though! A contribution to the docs to make it easier to find might be in order. It's in the help files but the README could use a mention. Would you like to do that?

uWynell:
> Umm, did you check the docs first (`:help NerdCommenter`)? I think this already exists:
> 
> ```vim
> let g:NERDCreateDefaultMappings = 0
> ```
> 
> Additionally all of the default mappings only actually get run if the same function isn't already bound to something else, so you can create custom bindings to some functions and skip the default ones while still getting the other defaults.

Yes, I will do that.

uWynell:
> Umm, did you check the docs first (`:help NerdCommenter`)? I think this already exists:
> 
> ```vim
> let g:NERDCreateDefaultMappings = 0
> ```
> 
> Additionally all of the default mappings only actually get run if the same function isn't already bound to something else, so you can create custom bindings to some functions and skip the default ones while still getting the other defaults.

I'm new to VIM, so its docs still look intimidating to me, I rather look at README.

Personally I just wanted to use <leader>c to toggle comment and nothing else, but if I did not remove the default mappings, VIM would have timeout after pressing <leader>c for the case I would add other keys.

alerque:
> Personally I just wanted to use `<leader>c` to toggle comment and nothing else, but if I did not remove the default mappings, VIM would have timeout after pressing `<leader>c` for the case I would add other keys.

Aaa yes that would be true, even though the actual thing you bound to a key wouldn't get rebound it would still be listening for other possible combinations starting with `<leader>c`. That's a good use case for not setting any bindings at all by default.

Issue #421, opened 2020-03-03
Make space delimiter an option on a per language basis


alerque:
Right now we have a global preference for using spaces after delimiters:

```vim
let g:NERDSpaceDelims = 1
```

I would like to overhaul this such that each language in our array of supported file types specifies a default integer number of spaces.

Example:

```vim
let s:delimiterMap = {
    \ 'aap': { 'left': '#', 'spaces': 1 },
  \}
```

This would make it easier to manage the cases where some language have stringent style guides, others even mandate spaces for proper lexing, while many people prefer no spaces at all. Additionally it keeps our current "uncomment" code resilient to variations in these preferences, something hard coding the space into the delimiter ruins.

See also #328, #278, #202, #183, #420.

alerque:
This idea can probably be expanded to other settings such as default alignment (see e.g. [this comment](https://github.com/preservim/nerdcommenter/issues/202#issuecomment-341849824) for a use case).

PR #447, opened 2020-12-02
NERDCommenterInvert will not work correctly if there isn't an extra space (Haskell)

attacker0211:
see this image:
image
after:
image


alerque:
There are known bad side effects of adding spaces to the comment delimiter string. Whether you are working in a language that requires them or just prefer them the current remedy is the same, set:

```vim
let g:NERDSpaceDelims = 1
```

See #421 and the plethora of related linked issues there. Haskell also came up specifically in #420.

PR #453, opened 2021-01-04
remove extra space

bkbncn:
Hi, there is extra space for python and pyrex, but not for the other languages. We should fix the inconsistence.


bkbncn:
All right, now I understand how come the inconsistence, and agree not everyone will customize this plugin, including me, I didn't customize before.

And my workaround about `python` is following, inspired by #202, hopefully help others.
```
let g:NERDSpaceDelims = 1
let g:NERDAltDelims_python = 1
```

Issue #315, opened 2017-12-03
Ability to commend fold


mitermayer:
It would be awesome if we could comment a fold without having to visually select lines.

alerque:
You probably can. Does vim not have motion or range commands for the current fold?

mitermayer:
Would be nice if that was part of nerdcommenter

alerque:
I'm unsure what you think this would look like. The way vim and plugins typically work is that each plugin handles some specific task and provides specific functions. Good plugins allow these functions to interact with other vim functions and the user can mix and match them as their workflow requires.

* NerdCommenter provides comment related functions.
* Vim  provides range, motion, selection, [text objects](https://blog.carbonfive.com/2011/10/17/vim-text-objects-the-definitive-guide/), etc.
* You can run any mode of `NERDComment()` on any of the above.

If this plugin were to try to get into pairing the above on it's own instead of leaving that to the user the possibilities would mushroom out of control really fast. The iterations of which comment style (line? block? sexy? minimal?) and motion type (word, sentence, paragraph, quoted string, fold, indent level, inside? outside?) would be in the thousands.

Instead the way vim works is you mix and match yourself. If a particular combination of motion and action is too cumbersome for you, you can make it more accessible with a mapping.

I assume you've read the `:help fold` documentation for how to move around and manipulate folds. If you want to operate on them a convenience text object for the current fold can be handy. For example of how to do this see [this tutorial](vim.wikia.com/wiki/Creating_new_text_objects).

The short version is you'll want these two lines in your rc file:

```vim
vnoremap af :<C-U>silent! normal! [zV]z<CR>
omap af :normal Vaf<CR>
```

Now you have text object `af` (a fold). Lets say you like the sexy comment style and want to have quick access to that comment style for the current fold:

```vim
map <leader>cf Vaf:call NERDComment(1, 'sexy')<Cr>
```

Issue #260, opened 2016-07-31
Commenting a motion


tstelzer:
Generally speaking *plugin X does Y* is not a good argument for implementing a feature, I know, however: I very much miss the ability to comment a motion, from tpopes commentary. Prepending a count helps in a way, but still lacks the precision I am used to in vim. Things like:

Comment out 3 lines up:
`gc3k`

... outer paragraph:
`gcap` 

... inner brackets
`gcib`

etc.

A workaround would be to select first, and comment out second, but a motion would be preferable.

Combining motions with the multitude of your plugins features would be brilliant indeed. Hope I am not beating a dead horse here, I haven't found a mention of motions in either the docs nor the issues.

*In either case, thanks for your great plugin, just found it and am very happy using it!*

alerque:
I agree this would be a great thing to add. Basically it would just amount to a parallel set of functions that took a motion as a required argument and passed that range to the various comment functions rather than the current selection. You could then bind those functions however you liked. For many scenarios this would just be a reversal of the key sequence already available and one less character (`comment_binding<motion>` instead of `v<motion><comment_binding>`), but for some scenarios it would enable more flexibility because motion commands are repeatable in a way that selections are often not.

That being said the original author isn't actively developing this at all right now and I'm basically on the project as a community maintainer. Given my limited time, I'm unlikely to sit down and implement this feature anytime soon. If anybody want's to take a stab at it I will be happy to review and merge when it works.

PR #454, opened 2021-01-11
Add a note on binds for text object motions

richchurcher:
Thought I'd drop this in the readme to explain how to mimic comment motions, _a la_ commentary.

See also #315, #260. Thanks for continuing to provide community support for nerdcommenter!


alerque (approved):

alerque:
Thanks for taking the time to contribute! I agree this isn't quite as good as having first-class motion support, but it's nice to have the documentation on-hand for what people that might want one or two equivalents for their favorite motions.

PR #446, opened 2020-12-01
Add support for Rego (Open Policy Agent)

priyanshsaxena:
[OPA](https://www.openpolicyagent.org/) [Rego](https://www.openpolicyagent.org/docs/latest/policy-language/)

PR #459, opened 2021-01-29
Update rego comment to #

jnsf-cg:
Update to the [Open Policy Agent ](https://www.openpolicyagent.org) comment character. Based on the [policy-language](https://www.openpolicyagent.org/docs/latest/policy-language/), a comment begins with `#`.

image

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -353,7 +353,7 @@ let s:delimiterMap = {
     \ 'rc': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'rebol': { 'left': ';' },
     \ 'registry': { 'left': ';' },
-    \ 'rego': { 'left': ';' },
+    \ 'rego': { 'left': '#' },
     \ 'remind': { 'left': '#' },
     \ 'renpy': { 'left': '# ' },
     \ 'resolv': { 'left': '#' },


alerque (commented):
Thanks for taking the time to contribute.

I'm unfamiliar with Rego, so extra thanks for the documentation links. I don't see anything noting when this changed happened or if the legacy (?) semicolon method is still valid. Should `;` be listed as an alternate syntax?

jnsf-cg:
I was poking around the opa code base and the [only token for the comment](https://github.com/open-policy-agent/opa/blob/b535cbfde11445df658878af9e8290ed38b9b4ab/ast/internal/scanner/scanner.go#L111-L113) appears to be `#`. I'm thinking that `;` was used early on and then removed.

When I go to the [OPA Rego Playground](https://play.openpolicyagent.org/) and try to use `;` as a comment, I get an error.

```rego
; hello
```

Output
```
1 error occurred: policy.rego:20: rego_parse_error: unexpected ; token
	; hello
	^
```

alerque:
> I'm thinking that `;` was used early on and then removed.

Early on? Is December of last year really such distant past? I remember checking the docs myself when I merged this in #446. Clearly the same docs linked then say something different now so I'm going to merge this, but color me just a bit confused.

@priyanshsaxena Any comments?

alerque (approved):

jnsf-cg:
Doh my 🧠  kept thinking that was a year ago. Hmmmm that is very odd.

PR #463, opened 2021-02-22
Update of UC4 Scripting

teaVeloper:
Hi,
I forgot to include the changes i made, so it behaves same way as in python. Adding a Space after the comment symbol. Makes scripts better readable.
Also included the 'leftAlt' in plain format. 
So basically same behavior as python, but different symbol.
Thanks!

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -450,7 +450,7 @@ let s:delimiterMap = {
     \ 'typescript': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'typescriptreact': { 'left': '//', 'leftAlt': '{/*', 'rightAlt': '*/}' },
     \ 'uc': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
-    \ 'uc4': { 'left': '!' },
+    \ 'uc4': { 'left': '! ', 'leftAlt': '!' },
     \ 'uil': { 'left': '!' },
     \ 'upstart': { 'left': '#' },
     \ 'vala': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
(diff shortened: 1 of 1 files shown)


alerque:
Including spaces in the delimiter (as opposed to turning on `let g:NERDSpaceDelims = 1` is a giant can of worms that causes all sorts of problems. We're trying to get *away* from this, not add more problem cases. The solution we're working towards for Python and others is in #421. See also the myriad of linked issues there.

I'd love to see a PR or some help implementing a per language preference array so this can get a proper resolution. I don't think I want to merge something that adds spaces to the base delimiter except in the case where it is required to parse the language. I fully agree spaces make comments more readable, but that isn't a justification to ruin the functionality for people that don't want them.

plugin/NERD_commenter.vim
```diff
@@ -450,6 +450,7 @@ let s:delimiterMap = {
     \ 'typescript': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'typescriptreact': { 'left': '//', 'leftAlt': '{/*', 'rightAlt': '*/}' },
     \ 'uc': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
+    \ 'uc4': { 'left': '! ', 'leftAlt': '!' },
```
alerque, line 453: Please drop the space from the base delimiter (and alternate). I understand the space will be proffered by many/most programmers (myself included), but if it isn't mandated my the language parser then we're going to leave that up to the user's preferences — for which there is a setting. I'm happy to merge the additionally language support, but for the special space treatment you should either use the space-delims setting or override this one language in your own RC file.

alerque (changes requested):

teaVeloper:
Hi,
Thank you a lot for your detailed answer! 
I didn't check the options carefully enough. But will do so in future.
I will close the PR as the filetype was already included in my last PR and so I guess everything is set fine.

Thanks for the support at contributing.
BR

PR #465, opened 2021-04-05
Add hocon filetype

GabrielNagy:
Add support for commenting hocon files.

[1] https://github.com/lightbend/config/blob/main/HOCON.md#comments

### DIFF
--- a/plugin/NERD_commenter.vim
+++ b/plugin/NERD_commenter.vim
@@ -204,6 +204,7 @@ let s:delimiterMap = {
     \ 'hbs': { 'left': '{{!-- ', 'right': ' --}}' },
     \ 'hercules': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'hive': { 'left': '-- ' },
+    \ 'hocon': { 'left': '//', 'leftAlt': '#' },
     \ 'hog': { 'left': '#' },
     \ 'hostsaccess': { 'left': '#' },
     \ 'htmlcheetah': { 'left': '##' },


alerque:
I don't know anything about hocon, but the docs you link to are ... misinformed.

> As in standard JSON and standard HOCON, any text between // or # and the next newline is considered a comment and is ignored (unless the // or # is inside a quoted string).

Standard JSON does not have a valid comment syntax at all. This is something various derivatives frequently change, but it's not encouraging that their documentation doesn't know where they are starting from.

plugin/NERD_commenter.vim
```diff
@@ -204,6 +204,7 @@ let s:delimiterMap = {
     \ 'hbs': { 'left': '{{!-- ', 'right': ' --}}' },
     \ 'hercules': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'hive': { 'left': '-- ' },
+    \ 'hocon': { 'left': '#' },
```
alerque, line 207: Why use `#` as the default? The docs you linked to used `//` in most of their examples, it seems like that might be more idiomatic. Whichever one you use, you should also include a `leftAlt` key with the other one.

alerque (changes requested):
Thanks for taking the time to contribute.

GabrielNagy:
> Standard JSON does not have a valid comment syntax at all. This is something various derivatives frequently change, but it's not encouraging that their documentation doesn't know where they are starting from.

I agree, I just clicked on the first google result. I think this is the correct source: https://github.com/lightbend/config/blob/main/HOCON.md#comments

I'll update the code.

plugin/NERD_commenter.vim
GabrielNagy: No reason in particular, just Ruby mnemonics. I updated the code to default on `//`.

alerque (approved):

PR #466, opened 2021-04-13
Add Directions for Vim 8+ Packages

natesire:
Added directions to install with Vim 8+ packages like nerd tree README says.


alerque (approved):
Thanks for taking the time to contribute. I rewrote your commit messages to reflect what was actually changed in each.

PR #469, opened 2021-05-28
Fix multibyte character bug in block comment

tomtomjhj:
<Plug>NERDCommenterComment using block comment on visual block whose
last char is unicode breaks up the unicode char. Example:

    a가 → /* a<ea> */<b0><80>

To fix this, add the character's byte length to the index used for
chopping the right part of the text block to be commented.


alerque (approved):
Thanks for taking the time to contribute!

PR #470, opened 2021-07-29
fix and refactor SetUpForNewFiletype

tomtomjhj:
* fix: variables are not set for a new buffer with filetype = s:last_filetype
* remove unused parameter and simplify name

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -447,31 +447,20 @@ endif
 " Section: Comment mapping functions, autocommands and commands
 " ============================================================================
 
-" Function: nerdcommenter#SetUpForNewFiletype(filetype) function
+" Function: nerdcommenter#SetUp() function
 " This function is responsible for setting up buffer scoped variables for the
-" given filetype.
-"
-" Args:
-"   -filetype: the filetype to set delimiters for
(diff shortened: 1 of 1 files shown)


alerque (approved):
Thanks for this contribution! I scratched my head at first but it really does look like it accomplished all the goals of the original in a *much* simpler way, so nice refactoring.

alerque:
By the way I don't know if you saw from a week or so ago I actually invited you to join this org to have a little more free hand in helping out (e.g. being able to close issues, review and even merge other people's PRs, merge your own PRs with just a review approval from another contributor that doesn't have to be me, etc.). Feel free to join, and know that I don't expect that to mean any obligation to contribute, just a few more tools at your disposal when you do feel like it.

tomtomjhj:
@alerque I was a bit hesitant to accept the invitation because I probably won't be able to actively monitor the issues and PRs. But at least I think I can review and merge PRs on nerdcommenter when requested.

BTW the invitation seems to be expired. Can you please send it again? Thanks.

alerque:
Sure, I guess invitations have a 7 day timeout. I just renewed it.

I fully understand if you can't/don't actively monitory all issues and PRs, but at least if occasion does come up again the barriers to contribution will be a little lower.

PR #475, opened 2021-07-31
minor refactoring

tomtomjhj:


alerque (approved):
Nice work. I tried to `git blame` to see where some of the completely unnecessary complex expressions came from and they predate the 2007 start of the Git repository.

PR #479, opened 2021-09-02
fix 'g:NERDAllowAnyVisualDelims' behavior

jiaobuzuji:
Hi Maintainer

I found that the behavior of the 'g:NERDAllowAnyVisualDelims' option is inconsistent with that described in the document. 

https://github.com/preservim/nerdcommenter/blob/2c87cae76ed88f3b0c957df68c03573ecafa3728/doc/nerdcommenter.txt#L524-L545

PR #440 .

Best Respect.


alerque (approved):
Thanks for taking the time to contribute. I had to read tho code about 5 times before the logic made sense but I think this change is correct. Since I don't actually use the feature much I have to wonder if we'll take anybody by surprised based on what the code *did* instead of how it was *documented*. I guess we'll see.

Issue #425, opened 2020-05-04
c/c++ alternative delims


cridemichel:
Dear developers,
first of all thank you very much for the very nice and useful plugin!
I have a minor issue to report. 
To reproduce the issue, do the following steps:
```list
- select in c/c++ the alternate delims '//' by using <leader>ca 
- select several lines in visual mode, e.g.
```
```c
printf("one\n");
printf("two\n");
printf("three\n");
```
```list
- comment these lines by using <leader>cc
```
what I get is the following:
```c
  /* printf("one\n"); */
  // printf("two\n");
  /* printf("three\n"); */
```
while I expected the following output:
```c
  // printf("one\n");
  // printf("two\n");
  // printf("three\n");
```

kind regards Cristiano

alerque:
I agree something doesn't look right there!

We'd be happy to accept a PR that fixed this if anybody wants to tackle it!

VishalNayek:
Hi! I am up for the challenge. You can assign this to me.

VishalNayek:
Can you help me on how to actually run the source code, I am a bit stuck!

alerque:
@VishalNayek Sure, but I'm not sure what you mean. If you have the plugin installed somewhere you should be able to just edit the plugin code wherever you already have it installed, then rerun vim to use your changes.

cridemichel:
Hi, any chance to have it fixed?

cridemichel:
it seems that a fix is rather easy, in file plugin/NERD_commenter.com
inside the function  s:CommentBlock, the following if...endif statement
```vim
 if !s:Multipart() && g:NERDAllowAnyVisualDelims && s:AltMultipart()
        let switchedDelims = 1
        call s:SwitchToAlternativeDelimiters(0)
    endif
```
should be replaced with
```vim
 if !s:Multipart() && !g:NERDAllowAnyVisualDelims && s:AltMultipart()
        let switchedDelims = 1
        call s:SwitchToAlternativeDelimiters(0)
    endif
```
since the switch to alternative delimiters there 
should occur only if g:NERDAllowAnyVisualDelims = 0

Do you need a pull request or you can make this minor change by yourself?

best Cristiano

alerque:
@cridemichel A PR would be easier for others to test!

cridemichel:
I have just created a PR
best C.

cridemichel:
Dear alerque,
is there any chance to have it merged into master?
best Cristiano

[...]

PR #480, opened 2021-11-10
fix for C/C++ alternative comments

cridemichel:
the issue reported in #425 is back and this is a possible fix

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -651,7 +651,7 @@ function! s:CommentBlock(top, bottom, lSide, rSide, forceNested) abort
     "alternative delimiters (if THEY are) as the comment will be better and more
     "accurate with multipart delimiters
     let switchedDelims = 0
-    if !s:Multipart() && g:NERDAllowAnyVisualDelims && s:AltMultipart()
+    if !s:Multipart() && !g:NERDAllowAnyVisualDelims && s:AltMultipart()
         let switchedDelims = 1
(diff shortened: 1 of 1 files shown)


alerque:
I think this merits some deeper review. We already implemented this fix in #440 and then had to revert it in #479 because it broke something else. I'm happy to see the issue fixed but we need to figure out the logic to do so without breaking the other feature. This specific "fix" works for the one problem, but it introduces another.

I'm on the road right now and unable to dig into it much farther, but before I merge this we need to confirm that it doesn't just break the same other thing all over again. I suspect a more complete set of logic is going to be needed to cover all possible combinations of features.

cridemichel:
Dear Caleb,
thank you very much for the clarification, I think that the real workaround to bug #425 is simply to respect 
the nerdcommenter actual logic, i.e. if g:NERDAllowAnyVisualDelims is set to 1 and one wants to force the use 
of alternative delimiters, one has to select test in visual mode linewise (V). Indeed, if I select test with V, alternative
delimiters are correctly used if selected with "<Leader>ca". You can close bug #425,
 
best Cristiano

Cristiano De Michele
***@***.***

amgodoi:
Any chance to get it merged?

xian-wen:
Anything update? When I run `<leader>ca`, the command line says "NERDCommenter:Now using //  to delimit comments". After that, however, when I run `<leader>cc` or `<leader>c<space>`, the code is still commented as /* <code> */.

BigPeet:
Maybe related to #536

PR #482, opened 2021-11-26
Simplify and fix NERDCommenterInsert

tomtomjhj:
* Simplify comment delimiter insertion.
* Remove the code that unnecessarily touches the leading spaces/tabs.
* Fix a bug in which `<Plug>NERDCommenterInsert` inserts the comment
  delimiters in a wrong position when `col('.')==1`.
  Example: `1|234` (cursor at `|`) → `/* */1234`.
* NOTE: Should avoid `feedkey(.., 'ni')` for inserting comment
  delimiters to ensure that `NERDCommenter_after` is called after
  inserting the delimiters. `feedkey` only adds the input to the queue,
  which will be processed after exiting the script context. On the other
  hand, `:normal` and `feedkey(.., 'x')` are eagerly processed.

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -1429,12 +1429,13 @@ function! s:PlaceDelimitersAndInsBetween() abort
     "    Otherwise, use 'i'.
     let insert = col('.') > strlen(getline('.')) ? 'a' : 'i'
     " 2. Insert comment delimiters.
-    " 3. Move the cursor to the left of the closing delimiter.
+    " 3. Move the cursor to the left of the closing delimiter, without
+    "    breaking undo sequence.
     " 4. Enter insert normal mode again without changing the cursor position.
     "    This ensures that returning to the insert mode after finishing the
     "    script execution does not move the cursor.
-    "                 ( 1  )   (    2     )   (                3               )   (      4     )
-    execute 'normal!' insert . left . right . repeat("\<Left>", strchars(right)) . "\<C-\>\<C-O>"
+    "                 ( 1  )   (    2     )   (                   3                   )   (      4     )
+    execute 'normal!' insert . left . right . repeat("\<C-G>U\<Left>", strchars(right)) . "\<C-\>\<C-O>"
 endfunction
 
 " Function: s:RemoveDelimiters(left, right, line)
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -1422,46 +1422,19 @@ function! s:PlaceDelimitersAndInsBetween() abort
     let left = s:Left({'space': 1})
     let right = s:Right({'space': 1})
 
-    let theLine = getline('.')
-    let lineHasLeadTabs = s:HasLeadingTabs(theLine) || (theLine =~# '^ *$' && !&expandtab)
-
-    "convert tabs to spaces and adjust the cursors column to take this into
-    "account
-    let untabbedCol = s:UntabbedCol(theLine, col('.'))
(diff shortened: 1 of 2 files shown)


autoload/nerdcommenter.vim
```diff
@@ -1419,46 +1419,19 @@ function! s:PlaceDelimitersAndInsBetween() abort
...
+    "    If the cursor is on EOL (right of the last char), use 'a'.
+    "    Otherwise, use 'i'.
+    let insert = col('.') > strlen(getline('.')) ? 'a' : 'i'
+    " 2. Insert comment delimiters.
+    " 3. Move the cursor to the left of the closing delimiter.
+    " 4. Enter insert normal mode again without changing the cursor position.
+    "    This ensures that returning to the insert mode after finishing the
+    "    script execution does not move the cursor.
+    "             ( 1  )   (    2     )   (                3               )   (      4     )
+    exe 'normal!' insert . left . right . repeat("\<Left>", strchars(right)) . "\<C-\>\<C-O>"
```
alerque, line 1434:
```suggestion
    execute 'normal!' insert . left . right . repeat("\<Left>", strchars(right)) . "\<C-\>\<C-O>"
```

Just for consistency when searching the code base I'd rather always spell out full command names.

plugin/nerdcommenter.vim
```diff
@@ -125,7 +125,7 @@ function! NERDCommentIsCharCommented(line, col)
     return nerdcommenter#IsCharCommented(a:line, a:col)
 endfunction
 
-inoremap <silent> <Plug>NERDCommenterInsert <Space><BS><Esc>:call nerdcommenter#Comment('i', "insert")<CR>
+inoremap <silent> <Plug>NERDCommenterInsert <C-\><C-O>:call nerdcommenter#Comment('i', "Insert")<CR>
```
alerque, line 128: Can you explain the change in key sequences here? I'm concerned this might have side effects like being affected by user `<Leader>` key settings.

alerque (changes requested):
Wow, thanks for working on this. As far as I can see from a code review this looks great. I love seeing bad code refactored into something simpler. I had one nitpick, but otherwise I think I'm good with this..

That being said I have not been able to test it on my system yet. I'm on the road with a tight schedule and not much time for fiddling. I would love to see if some 3rd party could give this a test drive and make sure it doesn't break anything in their workflows before I merge.

Any takers?

plugin/nerdcommenter.vim
tomtomjhj: `<C-\><C-O>` (`:h i_CTRL-\_CTRL-O`) enters the "Insert Normal mode" that lets you run a single normal mode command and return to insert mode. In this state, `:` (which is a normal mode command) enters the cmdline mode to run `call ...`. After `call ...` command finishes, vim will return to insert mode. 

The old version exits insert mode (`<Esc>`), and returns to insert mode using `startinsert!` or `feedkeys('a', 'ni')` (in `s:PlaceDelimitersAndInsBetween`). I'm not sure what `<Space><BS>` is for. It predates the history of this repository..

alerque: Thanks for the explanation. I am guessing `<Space><BS>` was intended as a hack to have an undo point, but whether that guess is correct or not I suspect it was and is the wrong solution to whatever problem it was meant to address.

alerque (approved):
LGTM. It would still be nice to have at least one other user test drive this and confirm it works as I haven't been able to yet. If I do get a chance I will, but anybody else giving this a spin would help!

PR #485, opened 2021-12-28
Add SetUp call in IsLineCommented function definition

ftfunjth:
After I add a custom keymap likes following segment
```
nmap <silent> <expr> gc nerdcommenter#IsLineCommented(line(".")) ?
 \ "\:call nerdcommenter#Comment(1, 'uncomment')<CR>"
 \ : "\:call nerdcommenter#Comment(1, 'alignBoth')<CR>"
```

Call the map in a buffer that without any nerdCommenter's comment function been invoked before, vim will complain "E121: Undefined variable: b:NERDCommenterDelims".

Add `nerdcommenter#SetUp` invoke will make vim stop complaining.

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -1153,6 +1153,7 @@ endfunction
 "   -lineNo:    the line number of the line to check
 " Return: Number, 1 if the line is a comment, 0 else
 function! nerdcommenter#IsLineCommented(lineNo) abort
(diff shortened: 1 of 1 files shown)


alerque (approved):
Thanks for taking the time to contribute this fix!

alerque:
Great first PR on GitHub by the way, you explained the problem and the need for a fix, included the fix, and even used a good commit message. Rock on.

PR #486, opened 2022-01-15
Add support for ATS and Lean

phlip9:
I've had these sitting in my local dotfiles for a while; figured I'd push them upstream : )

ATS: http://www.ats-lang.org/
Lean: https://leanprover.github.io/


alerque (approved):
Thanks for the contributions!

PR #487, opened 2022-02-12
Add support for Starkware cairo

codemedian:
Adding support for cairo language by Starkware: https://www.cairo-lang.org/

There isn't an official ft plugin for it yet, but this is what's being used by the community as of now with the ft being defined as `cairo`
https://github.com/starkware-libs/cairo-lang/tree/master/src/starkware/cairo/lang/ide/vim


alerque (approved):
Thanks for the contribution and docs to make it easy to review.

PR #488, opened 2022-02-15
fix: added a space to Haskell one-line commands

fyusuf-a:
For example, it transforms '&&' to '-- &&' so that it is not interpreted as the '--&&' infix operator


tomtomjhj (commented):
I think this should be configured with [`g:NERDSpaceDelims`](https://github.com/preservim/nerdcommenter/blob/2a0a05ff983aa62d74ba868aadf89deb93dd5454/doc/nerdcommenter.txt#L494)

Issue #17, opened 2010-12-06
Different comment style for file regions


splondike:
Would it be possible to have the comment style change depending on the cursor's position within a multi-language file? The specific case I'm interested in is Javascript and CSS embeded in a HTML document. Within the script tags i'd like to use // or /**/ style, /**/ within the style tags and  for the HTML:

```
<html>
<style>
    html{
        /*background:green;*/
    }
</style>
<script>
    //var foo = "bar";
</script>
<!--<body></body>-->
</html>
```

I had a look around and this doesn't seem to be possible in a 'nice' way (i.e. there isn't a 'current_region_code_type' variable set). I've never done any vim programming though, is this correct?

chadfurman:
<html>
...
<body>
<script>

    

</script>
</body>

Please re-open :)

mblarsen:
This seems to work well https://github.com/posva/vim-vue#how-can-i-use-nerdcommenter-in-vue-files, but would be nice with a generic solution

alerque:
Thanks for pointing out that implementation @mblarsen. That's exactly the kind of thing we're going to have to do to make this feature work. In fact I think it will have to be implemented somewhat differently for detecting some embedded language pairs. I guess what this issue needs is some sort of framework for adding such hook based checks to certain languages, e.g. in HTML files by default have hooks that look for JS and CSS blocks, in SILE files look for LUA blocks, in Markdown look for code blocks with language tags, etc.

mblarsen:
Couldn't the plugin look for the nearest language syntax instead of using file type?

alerque:
@mblarsen If you know of a universal way to do quick, reliable detection of languages across the board from small fragments of code, then by all means lets hear it. As far as I'm aware the Unicorns of Neverland still have a corner on that market.

Looking for syntax of specific languages that are likely to be found in specific other ones might just be feasible, but the reason file type settings exist in editors in the first place and why they are typically based on file extension is because there isn't a practical implementation of any other method.

Keep in mind that the fragments may be very small:

```html
<a href="<?php if (JS) { echo "javascript:myPopup('<p>Text</p>'); } else { echo '#'; } ?>" style="margin-left: 1em; <?= $link_css ?>">Link</a>
```

Think about all the places your cursor could be in that line and what would be involved in detecting the "nearest" language and trying to interpret a comment command. Even given these language pairs include explicit clues, untangling this kind of mess via syntax detection would require accurate syntax parsers for every language involved.

[...]

PR #489, opened 2022-02-18
MOD: support for multiple languages in one file, like *.vue

FuDesign2008:
add support for vue file


alerque:
You don't say much, but I can guess the closure is because this doesn't really accomplish the goal, no?

For context, see #17 for a long standing request (that we would love to see implemented) to handle embedded file types, but we don't have a working solution yet. It is theoretically possible (and other plugins have managed similar effects) but isn't easy to retrofit in this plugin. Quite open to it, just saying it will take a bit of refactoring...

PR #494, opened 2022-07-24
Do not add extra space after left delim in empty line

Masquue:
`s:CommentLines` currently add an extra space after left delim regardless if the line is empty.

This causes the following code (cursor position represented as `|`):

```c
{
|
}
```
if using command `V\cc`, to be formatted as (space represented as `_`):

```c
{
//_
}
```

This patch removes this extra trailing space on empty lines in `s:CommentLines`.


Masquue:
`'NERDTrimTrailingWhitespace'` already does this.

But the doc is in correct as it says
> 'NERDTrimTrailingWhitespace'
> Specifies if trailing whitespace should be deleted when **uncommenting**

PR #497, opened 2022-09-11
add tla

nano-o:


alerque:
* http://lamport.azurewebsites.net/tla/tla.html
* https://github.com/florentc/vim-tla

alerque (approved):

PR #502, opened 2022-10-07
Add support for Svelte

brainwo:
Svelte language


alerque (approved):
Thanks for the contribution.

matthewblott:
The problem here is Html is just one part of the Svelte component. JavaScript and CSS end up using Html comments with this fix.

alerque:
Yes, currently we only filter by filetype. There is a long standing feature request to handled nested blocks of other languages, but so far nobody has contributed a working version that handles that. For some pairs and with other plugins that set comment strings it can sometimes work out better to disable this entirely.

matthewblott:
Okay thanks for the quick response. I've just set the comments for Svelte to use JavaScript type comments. It's a hack but that's mostly what I need to comment / uncomment in a Svelte component.

PR #503, opened 2022-10-07
Add support for Justfile

brainwo:
Support for https://github.com/casey/just


alerque (changes requested):
These entries should not have trailing spaces. There is an option for adding (or not adding) trailing spaces, and adding them as part of the comment separator breaks some uncomment and comment detection functions break. We have a long standing issue for a new way to handle languages that absolutely require the space as part of the comment separator token.

brainwo:
`just` uses `#` for shebang also ([as shown here](https://github.com/casey/just/blob/master/justfile#L161)). For comments, they usually have a space. That is why I added a trailing space and I saw some entries in this file using trailing spaces.

I have updated this pull request and removing trailing spaces in other entries.

alerque:
> For comments, they usually have a space.

This misses the point. Many (perhaps even most) languages *traditionally* use spaces after comment delimiters. That's what this is for:

```vimscript
let g:NERDSpaceDelims = 1
```

Spaces are *not* a part of the justfile grammar: https://github.com/casey/just/blob/master/GRAMMAR.md#tokens

alerque (changes requested):
I think the removal on renpy and julia are *probably* both correct but please don't mix and match interests here. We have a commit that adds Justfile support and we've researched the grammar. Please make this PR just about that without slipping other things in. I don't have time to research and review the other two at the moment and there is no reason to hold up this PR for that. We can start a new PR to cover those changes and research them as we're able.

alerque (approved):

PR #504, opened 2022-10-28
Fix README

nella17:
replace NERDComment to nerdcommenter#Comment
ref: #473


alerque (approved):

alerque:
Thanks for taking the time to contribute to the benefit of future folks reading the docs!

PR #508, opened 2023-01-02
feat(dosbatch): change `dosbatch` delimiters

MeShootIn:
Change `left` and `leftAlt` delimiters to nestable `REM ` (extra space character is required) for security and compatibility reasons. Here is two (the shortest) of the many justifications:

- [tldr](https://stackoverflow.com/a/12407934)
- [generalized](https://stackoverflow.com/a/12408045)

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -91,7 +91,7 @@ let s:delimiterMap = {
     \ 'dns': { 'left': ';' },
     \ 'docbk': { 'left': '<!--', 'right': '-->' },
     \ 'dockerfile': { 'left': '#' },
-    \ 'dosbatch': { 'left': 'REM ', 'leftAlt': '::' },
+    \ 'dosbatch': { 'left': 'REM ', 'nested': 1, 'leftAlt': 'REM ', 'nestedAlt': 1 },
     \ 'dosini': { 'left': ';' },
     \ 'dot': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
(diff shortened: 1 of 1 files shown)


alerque:
Is the extra space actually *required* for the syntax to be valid? Including the space here is problematic since it might get doubled or not properly removed when uncommenting based on user settings. If it is actually part of the language spec (and not just a recommended style) then we can go ahead, just know that it doesn't cleanly solve the problem and forces some people to change their configurations to work around the other issues with this plugin not supporting spaces as part of delimiters.

MeShootIn:
@alerque
> Is the extra space actually required for the syntax to be valid?

Yes, otherwise there will be an error when executing the command (see example below).

Input:
```cmd
@echo off
REM echo commented
REMecho comment error
echo printed
```
Output:
```
"REMecho" is not recognized as an internal or external command,
operable program or batch file.
printed
```

> Including the space here is problematic since it might get doubled or not properly removed when uncommenting based on user settings.

However, without the extra space and with the plugin's default settings, uncommenting the line will leave the space at the beginning.

> If it is actually part of the language spec (and not just a recommended style)

The `REM` command is the only part of the [language](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/rem) dedicated directly to commenting.

> just know that it doesn't cleanly solve the problem and forces some people to change their configurations

Yes, may be. However, `REM` is better than `::` because due to `::` there is a chance to get a subtle, but possibly fatal, error during script execution.

alerque (approved):
Okay then, you talked me into it.

PR #509, opened 2023-01-25
Allow hooks to change filetype and comment deilmiters

sankhesh:
This change fixes the issue there custom hooks change filetype to change delimiters but nerdcommenter wouldn't respect the changes.

Fixes #479 and adds support towards #17 with custom hooks like in https://github.com/posva/vim-vue/blob/master/readme.md#nerdcommenter


alerque (approved):
Thanks for the contribution!

PR #511, opened 2023-02-05
Update starkware cairo comments to latest syntax

codemedian:
Starkware has released updates to the Cairo language where they replaced the `#` comments with `//` and in the upcoming 1.0 version, they'll support `//` as well as `/* */` equivalent to Rust. 

Cairo 1.0 repository for reference: https://github.com/starkware-libs/cairo


alerque (approved):

alerque:
Thanks for taking the time to contribute, and for the documentation so I can review it easily.

Yuck to a language completely changing like this, but that's users problem. I agree this plugin should default to the current supported upstream language spec, and anybody working with exclusively old code can add their own configs for the language.

PR #514, opened 2023-04-20
Prevent caching of buffer commentstring

### LINKED ISSUE #477: `commentstring` value is not updated after first usage in the buffer
[suy/vim-context-commentstring](https://github.com/suy/vim-context-commentstring) fixes the issue of **Lue in VimScript** and [this issue of **HTML**](https://github.com/preservim/nerdcommenter/issues/17) by changing the `commentstring` value, but another issue appeared after using this, I think **nerdcommenter** stores the first use of `commentstring` in the current buffer and never read it again:

https://user-images.githubusercontent.com/42011920/129219251-a0862097-5831-4f09-960b-e1a070a43999.mp4

_Originally posted by @MuhammadSawalhy in https://github.com/preservim/nerdcommenter/issues/17#issuecomment-897709359_

ppwwyyxx:
Fix #477

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -460,10 +460,6 @@ endif
 " This function is responsible for setting up buffer scoped variables for the
 " current buffer.
 function! nerdcommenter#SetUp() abort
-    if exists('b:NERDCommenterDelims')
-        return
-    endif
-
     let filetype = &filetype
 
     "for compound filetypes, if we don't know how to handle the full filetype


alerque (approved):

alerque:
Thanks for taking the time to contribute!

alerque:
I had to revert this change. I think I understand what you were trying to fix but it completely blew up the plugin so it didn't work at all for my setup. I reverted it so people wouldn't run into the same total breakage. Can we look into what that was about?

ppwwyyxx:
This works in my setup. How do I reproduce the breakage so I can look into it?

ppwwyyxx:
I did notice one potential issue of infinite recursion. It's fixed here https://github.com/preservim/nerdcommenter/commit/a3e82853089e0f7a14c3475ec997f1aa371f4210 but I'm not sure if that's the breakage you're seeing.

alerque:
That commit you linked seems to fix the issue I was having with it. Can you submit that as a new PR?

PR #516, opened 2023-04-26
Add support for VCL

ponkio-o:
Added vcl (Varnish Configuration Language)
See. https://www.varnish-software.com/developers/tutorials/varnish-configuration-language-vcl/#comments

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -421,6 +421,7 @@ let s:delimiterMap = {
     \ 'vala': { 'left': '//', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'vasp': { 'left': '!' },
     \ 'vb': { 'left': "'" },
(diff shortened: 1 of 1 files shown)


alerque (changes requested):
Thanks for the contribution. And I do appreciate the whitespace cleanup, but can we please not do that in unrelated commits / PRs? The commit for adding VCL support should *only* add VCL support. Whitespace cleanup commits should *only* change whitespace. Even if it is reviewable now, these add up to make `git blame` and other tooling very difficult to use in the future.

ponkio-o:
@alerque Thank you for your review! I just reverted the unrelated changes.
https://github.com/preservim/nerdcommenter/compare/2a425ef49439574c33fd0202b372f413454457a4..4709b6df1273da3b300c3e466ed4788494f22b8c

alerque:
Thanks! A separate PR with code style cleanup would be fine by the way.

PR #521, opened 2023-06-02
Provide argument for bufnr

serpent7776:
There's an error in vim 8.0:

```
E119: Not enough arguments for function: bufnr
E116: Invalid arguments for function getbufvar(bufnr(),'NERDCommenterFirstInit')
E15: Invalid expression: getbufvar(bufnr(),'NERDCommenterFirstInit')
```

In that version, `bufnr` requires an argument, so pass `"%"`, which means current buffer.

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -492,7 +492,7 @@ function! nerdcommenter#SetUp() abort
             endif
         endfor
         " if g:NERD_<filetype>_alt_style is defined, use the alternate style
-        let b:NERDCommenterFirstInit = getbufvar(bufnr(),'NERDCommenterFirstInit')
+        let b:NERDCommenterFirstInit = getbufvar(bufnr('%'),'NERDCommenterFirstInit')
         if exists('g:NERDAltDelims_'.filetype) && eval('g:NERDAltDelims_'.filetype) && !b:NERDCommenterFirstInit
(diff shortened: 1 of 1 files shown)


alerque (commented):
Will this cause an error in any older/never versions?

serpent7776:
I checked this on vim 8.0 and 8.2 and both versions work.

Documentation says `"" or "%" can be used for the current buffer` on both versions. For 8.2 the first argument is optional, but for 8.0 it isn't.

I think it should be safe for newer versions, unless vim broke compatibility here.
Unsure about versions older than 8.0.

autoload/nerdcommenter.vim
```diff
@@ -492,7 +492,7 @@ function! nerdcommenter#SetUp() abort
             endif
         endfor
         " if g:NERD_<filetype>_alt_style is defined, use the alternate style
-        let b:NERDCommenterFirstInit = getbufvar(bufnr(),'NERDCommenterFirstInit')
+        let b:NERDCommenterFirstInit = getbufvar(bufnr("%"),'NERDCommenterFirstInit')
```
alerque, line 495:
```suggestion
        let b:NERDCommenterFirstInit = getbufvar(bufnr('%'),'NERDCommenterFirstInit')
```

alerque (changes requested):
Need to pass lints...

alerque (approved):

PR #525, opened 2023-08-03
Add new delimiter for HCL files

someone-stole-my-name:
Adds support for raw HCL files, docs [here](https://developer.hashicorp.com/packer/docs/templates/hcl_templates/syntax#comments).


alerque (approved):
Thanks for the contribution...

PR #530, opened 2023-12-18
Add delimiter for pamenv

huangyxi:
> The "#" character at start of line (no space at front) can be used to mark this line as a comment line.
https://man7.org/linux/man-pages/man5/pam_env.conf.5.html

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -280,6 +280,7 @@ let s:delimiterMap = {
     \ 'ox': { 'left': '//' },
     \ 'paludis-use-conf': { 'left': '#' },
     \ 'pandoc': { 'left': '<!--', 'right': '-->' },
+    \ 'pamenv': { 'left': '#' },
     \ 'pascal': { 'left': '{', 'right': '}', 'leftAlt': '(*', 'rightAlt': '*)' },
     \ 'patran': { 'left': '$', 'leftAlt': '/*', 'rightAlt': '*/' },
     \ 'pcap': { 'left': '#' },


alerque:
This will only partially work because I don't think we have any accommodations for forcing comments to be at the start of a line. That might be worth opening a feature request to track as it may affect other formats as well. I can merge this but results will be a little bit inconsistent since most of the macros assume comments can start anywhere in a line.

huangyxi:
Thank you for your response and for identifying a generic issue in commenting.

I agree with the point that the current modification may not strictly follow the official document, potentially leading to inconsistency. However, in this specific situation, commenting whether at the start of a line doesn't actually affect the final result. Since the initial release of Linux-PAM on GitHub in June 2000, leading white spaces have been skipped before the comment sign `#` detection. And I have confirmed this on an Ubuntu machine.

```c
	// `/modules/pam_env/pam_env.c`

	/* skip leading white space */
	key += strspn(key, " \n\t");

	/* skip blanks lines and comments */
	if (!key || key[0] == '#')
	    continue;
```
[initial commit of above lines in linux-pam](https://github.com/linux-pam/linux-pam/blob/ea488580c42e8918445a945484de3c8a5addc761/modules/pam_env/pam_env.c#L228-L233)
[current commit of above lines in linux-pam](https://github.com/linux-pam/linux-pam/blob/cf492d0cedf10d4128c5cf211998dff2260acb6a/modules/pam_env/pam_env.c#L937-L942)

Anyway, I understand and respect the decision, whether it involves merging this pull request or not.

alerque (approved):

PR #536, opened 2024-08-22
fix: correct logic for switching alt delimiters

carlostobon:
- Fixed a problem where toggling delimiters was not working due to copy statement.
- Confirmed that toggling works correctly across different buffers.

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -486,7 +486,7 @@ function! nerdcommenter#SetUp() abort
     let b:NERDSexyComMarker = ''
 
     if has_key(s:delimiterMap, filetype)
-        let b:NERDCommenterDelims = copy(s:delimiterMap[filetype])
+        let b:NERDCommenterDelims = s:delimiterMap[filetype]
         for i in ['left', 'leftAlt', 'right', 'rightAlt']
             if !has_key(b:NERDCommenterDelims, i)
                 let b:NERDCommenterDelims[i] = ''


alerque:
This is a direct revert of fea637c from #531 by @svalaskevicius. There must me something one or both of you are not testing because it can't be right both ways. What versions of VIM are you using to test with?

carlostobon:
Yeah it's essentially a revert of. Running archlinux with binary from the pkg manager and current version VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Aug 11 2024 20:40:17). Im probably doing something wrong but certainly that line is causing the bug.

alerque:
It looks like a pretty safe bet that @svalaskevicius is running NeoVIM based on his dotfiles:

https://github.com/svalaskevicius/dotfiles/blob/master/.config/nvim/conf/plugins.vim#L25

That might have something to do with why you're coming to opposite conclusions. I don't think we should just revert either way or even just gate it on VIM vs. NeoVIM, there is probably an actually right way to do this that works properly in both environments. That's what we should go for.

carlostobon:
Completely agree with you. Thanks for your time.

alerque:
(Personally I'd prefer to keep this open to track the known problem until we find a fix, the "closed" label doesn't encourage participation or responses which we kind of need here to sort this out.)

carlostobon:
Sounds perfect!

anthdono:
Confirming commit 3f860f2 breaks alternate delims in the latest Neovim nightly as of writing, however it is working fine in Vim 9.1.

I agree with @alerque, as this is a Vim library, it may not be appropriate to revert the change due to behaviour in Neovim, albeit a nuisance.

I'll try find some time to investigate a solution for #531 that works for both Vim and Neovim.

GDivino:
+1 on merging this PR, the toggling of alternative delimiters stopped working for me as well. Would be really nice to have this fix ASAP :)

scottchiefbaker:
I like my comments in C/C++ to be `//` so I use alternate delimiters. I'm also a frequent user of `NERDCommenterToggle`. When I toggle in a C file the first time it uses `//` and then every time after that it uses `/* ... */`. Applying this pull-request to the nerdcommenter code solves this problem for me.

This gets a +1 to merge from me.

I'm running Vim 8.0.1763 on Linux.

BigPeet:
> I like my comments in C/C++ to be `//` so I use alternate delimiters. I'm also a frequent user of `NERDCommenterToggle`. When I toggle in a C file the first time it uses `//` and then every time after that it uses `/* ... */`. Applying this pull-request to the nerdcommenter code solves this problem for me.
> 
> This gets a +1 to merge from me.
> 
> I'm running Vim 8.0.1763 on Linux.

I'm experiencing the exact same issue on Vim 9.1.1081. Applying this fix locally resolves the issue.

xdesai:
A couple more data points in case it's helpful. Although I'll admit I'm mostly restating what was said before.


In my regular vim
```
vim --version
VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Mar  7 2025 22:26:51)
macOS version - arm64
```
things currently work as expected running the latest master. If I apply this patch they actually stop working.


In my neovim 
```
nvim --version
NVIM v0.11.0
Build type: Release
LuaJIT 2.1.1741730670
Run "nvim -V1 -v" for more info
```
things do not work on the current master. But when I apply the above patch locally they do work again.

So, I do believe this PR is not safe to merge in as written. At the very least it probably needs a nvim/vim compat switch. Or maybe something more elegant like was mentioned above.

carlostobon:
I just compiled both Vim and Neovim from source in two different containers using the same GCC version to try to reproduce the issue.

Vim case:
`
-> gcc::latest image tag with gcc (GCC) 15.1.0
-> vim --version
    VIM - Vi IMproved 9.1 (2024 Jan 02, compiled May  2 2025 16:07:41)
-> vim-plug and nerdcommenter (latest master)
`

Nvim case:
`
-> gcc::latest image tag with gcc (GCC) 15.1.0
-> nvim --version
    NVIM v0.11.1
    Build type: RelWithDebInfo
    LuaJIT 2.1.1741730670
-> vim-plug and nerdcommenter (latest master)
`

Host system as reference:
`
-> OS: Gentoo Linux x86_64
-> Kernel: 6.13.5-gentoo
-> CPU: Intel Xeon E5-2697 v3 (28) @ 3.600GHz
`

In both Vim and Neovim, the issue with switching delimiters persists until I apply this commit.

Unless there's something specific about your architecture (which I think is unlikely), it seems more probable that there’s something else we're missing or the commit genuinely fixes the issue and should be merged.

carlostobon:
I just compiled both Vim and Neovim from source in two different containers using the same GCC version to try to reproduce the issue.

Vim case:
```
-> gcc::latest image tag with gcc (GCC) 15.1.0
-> vim --version
    VIM - Vi IMproved 9.1 (2024 Jan 02, compiled May  2 2025 16:07:41)
-> vim-plug and nerdcommenter (latest master)
```

Nvim case:
```
-> gcc::latest image tag with gcc (GCC) 15.1.0
-> nvim --version
    NVIM v0.11.1
    Build type: RelWithDebInfo
    LuaJIT 2.1.1741730670
-> vim-plug and nerdcommenter (latest master)
```

Host system as reference:
```
-> OS: Gentoo Linux x86_64
-> Kernel: 6.13.5-gentoo
-> CPU: Intel Xeon E5-2697 v3 (28) @ 3.600GHz
```

In both Vim and Neovim, the issue with switching delimiters persists until I apply this commit.

Unless there's something specific about your architecture (which I think is unlikely), it seems more probable that there’s something else we're missing or the commit genuinely fixes the issue and should be merged.

xdesai:
oh that's so interesting!
Your test is a much more clean room environment that my test was. It seems likely then that I've got some other plugin or vimrc setting that's messing with things, but I don't know what it could be :/

Issue #435, opened 2020-09-11
[Feature request] NERDCommenterYankAndPaste command


nwaywood:
A command that in addition to yanking (like the `NERDCommenterYank`) command also pastes the yanked code below the commented out version. 

For example:

Before running `NERDCommenterYankAndPaste` command
```
some 
code
block
```

After running `NERDCommenterYankAndPaste` command with the three lines selected in visual mode
```
// some
// code
// block
some
code
block
```

Obviously is you are in normal mode and not visual mode then it would just yank, comment and paste the current line.

I hope you like the idea!

alerque:
This sounds pretty workflow specific, and all the pieces are there already. Why not just add yourself a map that first calls the `Yank` function, then pastes it? This is what VIM is really good at: composability. VIM and plugins provide building blocks of specific actions, you compose them into whatever sequences you want to run.

nwaywood:
@alerque Thanks for the suggestion :)

I tried that already but without much success. Please note that I am not a Vimscript expert so I apologize in advance if I'm doing something stupid.

I wrote the following code in my `init.vim` to try and do what I want:

```
nnoremap gC :call Test()<cr>
function! Test()
   :call NERDComment('n', 'yank')
   exec 'normal! p'
endfunc

xnoremap gC :call Test2()<cr>
function! Test2()
   :call NERDComment('x', 'yank')
   exec 'normal! p'
endfunc
```

The first function/binding for normal mode works correctly but the function/binding for visual mode doesn't work at all.  Any idea what I'm doing wrong? Thanks

alerque:
For the visual mode example, before running the normal mode command I think you need to jump the cursor to the end of the visual selection and switch to normal mode.

nwaywood:
@alerque Yes I think that would if it wasn't for the other issue that I mentioned. If I modify the function so that it jumps to the end of the visual selection like this:

```vim
xnoremap gC :call Test2()<cr>
function! Test2()
   :call NERDComment('x', 'yank')
   exec "normal! '>"
   exec 'normal! p'
endfunc
```

It results in the following code. It seems like the `NERDComment('x', 'yank')` is doing something funky.

```
"""some
"""code
"""block
""some
""code
""block
"some
"code
"block
some
code
block
```

Any idea why this would be happening?

nwaywood:
@alerque I did some more testing and this code snippet

```vim
xnoremap gC :call Test2()<cr>
function! Test2()
[...]

PR #544, opened 2025-05-06
Add ability to specify register when yanking

### LINKED ISSUE #543: [Feature request] ability to specify register when yanking
`NERDCommenterYank` yanks to the default unnamed register `""`. Like all other yanking/pasting in vim, this is a good default, but it's also nice to be able to choose a register. One use case is creating a “comment and duplicate” function/mapping without clobbering the default register, as discussed in #435.

I don't know the internals or the right syntax, but from a UX perspective, I would hope for all of the following:

- `["x][count]<leader>cy` - Support standard register specification for NERDCommenterYank mapping, as with `["x]yy` and friends.
- `["x]<plug>NERDCommenterYank`, `<plug>NERDCommenterYank({register})`, or similar way to specify a register when invoking this way (e.g. in a user-defined function or mapping).
- `NERDComment({mode}, {type}, [{register}])` - Specify register when `type=Yank`.
- `let g:NerdYankRegister = x` - Set default register used by NERDCommenterYank.

jacktose:
`NERDCommenterYank` (`\cy`) yanks to the default unnamed register `"`. Like all other yanking, this is a good default, this gives the user an option to choose another register.
One use case is creating a “comment and duplicate” function/mapping without clobbering the default register, as discussed in #435.
Closes #543

## Goals
- ✅ `["x][count]\cy` - Support standard register specification for NERDCommenterYank mapping, as with `["x]yy` and friends.
- ❌ `["x]<plug>NERDCommenterYank`, `<plug>NERDCommenterYank({register})`, or similar way to specify a register when invoking this way (e.g. in a user-defined function or mapping).
  - I'm not sure if this is possible. IMO it's not critical because it can be accomplished by remapping to the default maps: `nmap \abc "x\cy`
- ❌ `NERDComment({mode}, {type}, [{register}])` - Specify register when `type=Yank`.
  - I'm very reluctant to mess with the main function signature.
- ✅ `let g:NERDDefaultRegister = x` - Set default register used by NERDCommenterYank.
  - This works. It defaults to `"` (so no behavior change by default) and is overridden when specifying a register with `"x`.

# To do
- [ ] Update documentation

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -1291,11 +1291,11 @@ function! nerdcommenter#Comment(mode, type) range abort
 
     elseif a:type ==? 'Yank'
         if isVisual
-            normal! gvy
+            execute 'normal! gv"'. v:register . 'y'
         elseif countWasGiven
-            execute firstLine .','. lastLine .'yank'
+            execute firstLine .','. lastLine .'yank '. v:register
         else
-            normal! yy
+            execute 'normal! "'. v:register .'yy'
         endif
         execute firstLine .','. lastLine .'call nerdcommenter#Comment("'. a:mode .'", "Comment")'
     endif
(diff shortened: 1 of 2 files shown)


jacktose:
@alerque Could you take a look?

Issue #329, opened 2018-04-01
change maker symbol


ghost:
Hi, is there any way to change the default marker symbol ?
Ex : `*` to `**` in C

alerque:
Please see `:help nerdcommenter` or more specifically in this case `:help NERDCustomDelimiters`.

albfan:
Looks like marker con sexy comments are selected using last char on longest left comment. In case of C comments is directly set as '*'

so as an example, 

```
let g:NERDCustomDelimiters = {
        \ 'xml': { 'left': '<!--', 'right': '-->', 'leftAlt': '<!-- ', 'rightAlt': '-->' },
        \ 'html': { 'left': '<!--', 'right': '-->', 'leftAlt': '<!--', 'rightAlt': '-->' }
    \ }
```

without set alt delimns for xml and html, you can use this trick to get an space as marker.

Still that's undesired, and one might want to have an empty marker.

Could make sense to allow:

```
let g:NERDCustomDelimiters = {
        \ 'xml': { 'left': '<!--', 'right': '-->', 'marker': ''},
        \ 'html': { 'left': '<!--', 'right': '-->', 'marker': ''},
    \ }
```

so user explicitly set the marker?

alerque:
If you have something to pose as a question or feature request or issue report, open a new issue @albafan.

albfan:
Looks to me the issue was valid and not resolved. I just try to complete an already existing issue so others can find it. #546 should fix it although I need to add 'nested' to avoid  an error plugin logic detects

```
let g:NERDCustomDelimiters = {
        \ 'c': { 'left': '/*', 'right': '*/', 'marker': '**', 'nested': 1}
    \ }
```

I'm not sure how nerdcommenter decides if use '//' or '/* */' my expectations would be to only set

```
let g:NERDCustomDelimiters = {
        \ 'c': {  'marker': '**'}
    \ }
```

PR #546, opened 2025-12-01
feat: Explicitly set marker for sexy comments

albfan:
Instead of extract the marker from last char of longest left comment allow to explicity set it

relates with #329

### DIFF
--- a/autoload/nerdcommenter.vim
+++ b/autoload/nerdcommenter.vim
@@ -490,6 +490,10 @@ function! nerdcommenter#SetUp() abort
 
     if has_key(s:delimiterMap, filetype)
         let b:NERDCommenterDelims = copy(s:delimiterMap[filetype])
+        if has_key(b:NERDCommenterDelims, 'marker')
+            let b:NERDSexyComMarker = b:NERDCommenterDelims['marker']
+            let b:NERDSexyComMarkerSet = 1
+        endif
         for i in ['left', 'leftAlt', 'right', 'rightAlt']
             if !has_key(b:NERDCommenterDelims, i)
                 let b:NERDCommenterDelims[i] = ''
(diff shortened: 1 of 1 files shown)


albfan:
to turn:

```
  <metadata>
    <property name="variant">dark</property>
    <property name="light-variant">Adwaita</property>
  </metadata>
```

into

```
  <!-- <metadata>
    <property name="variant">dark</property>
    <property name="light-variant">Adwaita</property>
  </metadata> -->
```

instead of

```
  <!-- <metadata>
  -   <property name="variant">dark</property>
  -   <property name="light-variant">Adwaita</property>
  - </metadata> -->
```

using:

```
let g:NERDCustomDelimiters = {
        \ 'xml':  { 'left': '<!--', 'right': '-->', 'marker': '', 'nested': 1 }
    \ }
```

albfan:
NOTE1: 'nested' is required to avoid an error:

            call s:NerdEcho('Sexy comment aborted. Nested sexy cannot be nested', 0)

NOTE2:
Expectation was to only need to set:
```
let g:NERDCustomDelimiters = {
        \ 'xml':  { 'marker': '' }
    \ }
```

But that doesn't do anything using `.cs`

autoload/nerdcommenter.vim
```diff
@@ -2242,7 +2246,7 @@ function! s:GetSexyComMarker(space, esc) abort
     let sexyComMarker = b:NERDSexyComMarker
 
     "if there is no hardcoded marker then we find one
-    if sexyComMarker ==# ''
+    if b:NERDSexyComMarkerSet == 0 && sexyComMarker ==# ''
```
alerque, line 2249: It doesn't look like this buffer value is getting set unconditionally, so this will error in the event it is not set ... no?

alerque (changes requested):
Resolve case where marker is *not* set.

